Issue #720, opened 2024-12-05
Comparison of virtual and in-person simulations for sepsis and trauma resuscitation training in Singapore: a randomized controlled trial


joshgilbert1994:
https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/KHAYXN

RCT, has item level data, too small for my study

Public Domain

curious-frances:
@ben-domingue @joshgilbert1994 Let me know what you guys think of the data if I need to make anyother modifications, otherwise I'll proceed with the rest of the components (pr + updating the tables)

[jian-wen-low-sepsis-and-trauma-resuscitation-2024.csv](https://github.com/user-attachments/files/24359024/jian-wen-low-sepsis-and-trauma-resuscitation-2024.csv)

ben-domingue:
i think it is fine. proceed! 
one question: just confirming that there should be `resp=0.5` values, yes?

curious-frances:
@ben-domingue yes, there are lots of values for resp=0.5!

curious-frances:
@ben-domingue Done with the rest of the components!

PR #1250, opened 2026-01-04
Comparison of virtual and in-person simulations for sepsis and trauma resuscitation training in Singapore: a randomized controlled trial #720

curious-frances:
@ben-domingue


ben-domingue:
@joshgilbert1994 might want to check this one out, see if it meets your rct requirements?

Issue #1470, opened 2026-06-02
Do People Across the World Want to Remember Positive Ingroup Histories?


ben-domingue:
https://osf.io/3w6ap/overview
https://psycnet.apa.org/fulltext/2027-22187-001.html

aanyatash:
Hi @ben-domingue, I worked on this dataset. I added the tags/dict/PR too.

I split it into 7 IRW tables – one for the participant-level scales, six for the event-level ratings:

- kazarovytska_2026_ingroup_scales has 13 items: national identification (×3), collective narcissism (×4), historical continuity (×3), PIVO (×3). 1–7 Likert.
- kazarovytska_2026_ingroup_event_srr is Social Relevance of Remembrance, 7 items
- kazarovytska_2026_ingroup_event_prr is Personal Relevance of Remembrance, 2 items
- kazarovytska_2026_ingroup_event_threat is symbolic + realistic threat, 4 items
- kazarovytska_2026_ingroup_event_valence is positivity/negativity of event + outcomes, 4 items
- kazarovytska_2026_ingroup_event_attribution is morality / agency / responsibility, 3 items.
- kazarovytska_2026_ingroup_event_presence is presence in media/politics, 2 items

[kazarovytska_2026_ingroup.zip](https://github.com/user-attachments/files/29355278/kazarovytska_2026_ingroup.zip)

ben-domingue:
for kazarovytska_2026_ingroup_scales, are the people rating the countries on something? in that case, i would have: 
- item=country
- rater=people doing rating (current `id`)

@aanyatash

ben-domingue:
@aanyatash kicking this one back around

aanyatash:
Hi @ben-domingue, for this table, each participant is from one country and rated only their own country/ingroup on 13 psychological Likert items (e.g. "The fact that I am [American] is an important part of my identity"). Country is not a variable being rated by multiple people, so that's why I left country as `cov_country` rather than `item`, and the items are the 13 Likert questions. Please let me know if this format makes sense – thank you!

Will be catching up on closing my open issues today and tomorrow and then start doing some new ones. Thanks!

ben-domingue:
hm. maybe we need `id=country`?

aanyatash:
Hi @ben-domingue, I updated the scales table to have that format. Please let me know what you think. I think I would have to change the other tables too to have id be the event and the rater be the participant to keep it consistent. Thanks!

[...]

PR #1509, opened 2026-06-25
Do People Across the World Want to Remember Positive Ingroup Historie…

aanyatash:
…s? #1470

### DIFF
--- a/data/kazarovytska_2026_ingroup.py
+++ b/data/kazarovytska_2026_ingroup.py
@@ -1,0 +1,260 @@
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import pandas as pd
+
+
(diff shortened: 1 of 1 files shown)


ben-domingue:
**Open 69 days, alongside #1512.**

The new `data/` contract check on `data/kazarovytska_2026_ingroup.py`: names and paths fine. One finding —

```
no_validator  writes a table (line 146) without calling the validator
```

Same one-line fix as #1512, and the same caveat: **776 of 843 existing scripts do not call it either**, so the rule is new.

Also unreviewed since 2026-06-25. @aanyatash, still wanted?
---

aanyatash:
Hi @ben-domingue, I updated the issue earlier this week, so it should be good to merge once reviewed – thank you!

PR #1517, opened 2026-07-02
covid19risktool_peters_2025

rubinastha:


ben-domingue:

`covid19risktool_peters_2025.R` sits at the top level rather than in `data/`, so it is outside the directory every other conversion script uses and outside what the new CI checks.

Moving it to `data/covid19risktool_peters_2025.R` is enough to get it reviewed mechanically.

Open since 2026-07-02 with no comments.
---

Issue #1479, opened 2026-06-10
Survey data on cognitive and emotional responses to visual anthropomorphism in health media


ben-domingue:
DOI      │ 10.6084/m9.figshare.32287584.v2 
 URL      │ https://figshare.com/articles/dataset/32287584        

double check license

tylermatteson22:
Looks safe. From the link: CC BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

However, everything is in Russian. Should I translate into English, or just translate column names locally for cleaning but write the final csv with all Russian text?

@ben-domingue @saviranadela

PS i dont see a way to assign myself to this issue.

ben-domingue:
@tylermatteson22 see email thread with subject "Non English data". i'm hoping that provides a fwd direction?

tylermatteson22:
#---------------------------------------------------------------------------
#Survey on cognitive/emotional responses to visual anthropomorphism in health media 
#All text is in Russian. 
#---------------------------------------------------------------------------

setwd(dirname(rstudioapi::getActiveDocumentContext()$path))
rm(list = ls())
library(tidyverse)

raw <- read_csv(
  "Survey data on cognitive and emotional responses to visual anthropomorphism in health media.csv",
  show_col_types = FALSE
)

#Keep the original Russian headers for item_text, then rename to short handles.
q_text <- names(raw) %>% str_replace_all("[\\r\\n]+", " ") %>% str_squish()
names(raw) <- c("timestamp", "age", "gender", "med_edu",
                "q04", "q05", "q06", "q07", "q08", "q09", "q10",
                "q11", "q12", "q13", "q14", "q15", "q16", "q17", "q18")
names(q_text) <- names(raw)

#Reformat Timestamp to Unix
df <- raw %>%
  mutate(
    id   = row_number(),
    date = as.numeric(as.POSIXct(timestamp, format = "%d.%m.%Y %H:%M:%S", tz = "UTC"))
  )

#Covariates
cov <- df %>%
  select(id, date,
         cov_age     = age,       # ordinal age bands
         cov_gender  = gender,    # Женский / Мужской
         cov_med_edu = med_edu)   # Да / Нет (medical education)

#--- helpers ----------------------------------------------------------------

#Likert: an already-numeric 1-5 rating -> one ordinal item. No nominal `text`.
make_likert <- function(data, col, item_id) {
  data %>%
    transmute(id,
              item        = item_id,
              item_family = NA_character_,
[...]

PR #1539, opened 2026-07-22
Adding table for Issue #1479

tylermatteson22:
Survey data on cognitive and emotional responses to visual anthropomorphism in health media #1479


ben-domingue:

`anthropomorphism_health_cleaning.R` sits at the top level. Every conversion script lives in `data/`, and the new CI only checks that directory — so as it stands this gets no mechanical review.

Moving it to `data/` is the first step. The naming convention there is `authorname_year_construct`, which also makes it findable next to its dictionary row.

Open since 2026-07-22 with no comments at all — that is on the review queue, not on you. Thank you for the contribution, and sorry it sat.
---

PR #1555, opened 2026-07-28
Create autobiologicalinterview.py

rubinastha:


ben-domingue:

`autobiologicalinterview.py` sits at the top level, so it is outside where every other conversion script lives — and outside what the new `data/` CI checks, which means it gets no mechanical review at all.

Moving it to `data/autobiologicalinterview.py` fixes both. Worth a look at the name too: the convention is `authorname_year_construct`, so something like `<author>_<year>_autobiographical_interview`. (Note the current spelling is `autobiological`, which looks like a typo for `autobiographical`.)

The PR body is empty as well — a reviewer needs the source, its license, and which issue it closes before anything else.

Open since 2026-07-28 with no comments, which is a queue problem rather than a problem with the work.
---

PR #1589, opened 2026-08-11
Sync irw-auto-itemtext skill: instructions/section_prompt boundary + two prior fixes

ben-domingue:
## Summary

Source-commit reference: this itemtext.qmd-derived change tracks
**datapages/irw commit `3da46a86662018e6f9bf09cfd89ee4ac51c6e947`** (PR #97,
"Clarify instructions vs section_prompt boundary in item-text standard",
merged into `datapages/irw:main` 2026-08-11), which inserted the following
rule into `itemtext.qmd`:

> `instructions` and `section_prompt` are scoped differently: `instructions`
> applies to the entire table regardless of `section_id`; `section_prompt`
> applies only to the items sharing one or more specific `section_id`
> values. The same span of source text should never be recorded in both
> fields. If framing or task-level text applies across the whole table,
> record it once in `instructions`. If it is specific to a subset of items
> sharing a `section_id` (e.g. a passage or context given before a testlet),
> record it in `section_prompt` only, even if it superficially resembles
> instructional language.

This PR brings `itemtext/.claude/skills/irw-auto-itemtext/` in sync with
that standard, and folds in two previously-flagged gaps in the same skill
that were confirmed (via `git log` + reading the current files) to not yet
be present anywhere in the repo.

- `references/itemtext_standard.md`: added the boundary rule as a short
  prose note directly below the `| Field | Definition |` table (it spans
  two fields so doesn't fit one row), and updated the header provenance
  note from the stale "(2026-07-27)" date to today's date plus the source
  commit/PR (`3da46a86662018e6f9bf09cfd89ee4ac51c6e947` / PR #97).
- `SKILL.md` Step 4: the `instrument/instructions` and
  `section_id/section_prompt` bullets already paraphrase each field's role;
  restated the same boundary rule there so no ambiguous/un-synced copy is
  left in the skill.
- `scripts/validate_items.R`: added a "Bare-integer item check" section
  (Fix 1) that flags tables whose `item` values are bare integers and
  reminds the runner that item-set/resp-range matching alone doesn't verify
  the reconstructed item-to-text mapping is correct.
- `SKILL.md` Step 4: added matching guidance for bare-integer item
  reconstruction (Fix 1) and a "Match the source's terseness" note (Fix 2)
  so generated text mirrors the source's brevity instead of adding
  explanatory boilerplate.

## Test plan

- [ ] `Rscript itemtext/.claude/skills/irw-auto-itemtext/scripts/validate_items.R <table> <items_csv>` still runs and now prints a "Bare-integer item check" section before "Resp set check".
- [ ] Read `references/itemtext_standard.md` and `SKILL.md` Step 4 to confirm the new prose reads naturally and doesn't duplicate/conflict with existing bullets.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

https://claude.ai/code/session_01NvrbrY1BgjUpfcqVLFtzVo

### DIFF
--- a/itemtext/.claude/skills/irw-auto-itemtext/SKILL.md
+++ b/itemtext/.claude/skills/irw-auto-itemtext/SKILL.md
@@ -101,14 +101,31 @@ Build the 4-tab structure (see `references/itemtext_standard.md` for exact colum
 in memory or as a scratch CSV — you don't have to actually create Sheets tabs, just
 produce data shaped like them before merging:
 
-- **instrument/instructions** — full instrument name + literal instructions text.
-- **section_id/section_prompt** — only for testlets/shared-passage items. If the
-  instrument has no such grouping, still emit one `section_id` per item (e.g.
-  `<table>_1`) with a blank `section_prompt`, rather than omitting the column — the
-  merge step needs a join key.
+- **instrument/instructions** — full instrument name + literal instructions text that
+  applies to the entire table regardless of `section_id`.
+- **section_id/section_prompt** — only for testlets/shared-passage items, and scoped to
+  just the items sharing that `section_id` (e.g. a passage or context given before a
+  testlet). If the instrument has no such grouping, still emit one `section_id` per item
+  (e.g. `<table>_1`) with a blank `section_prompt`, rather than omitting the column — the
+  merge step needs a join key. Never record the same span of source text in both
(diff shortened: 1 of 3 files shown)


ben-domingue (commented):
looks good

PR #1697, opened 2026-08-29
Use documented evidence for response-scale QC failures

sinew-07:
Unequal observed item ranges can currently produce a raw `run_qc()` failure on legitimate mixed-format tests. For example, 20 multiple-choice items scored 0/1 plus five constructed-response items scored 0–3 can fail, while reversing the majority can avoid that failure. This change makes width-only findings warnings and reserves response-scale failures for explicitly supplied source evidence.

The implementation lives in the shared `irw_validate` package, preserving the existing triage import. The branch incorporates `main` at `0aea245db300af0e9bb92b50952015e558a70107`, including the newer literal-response-token checks.

- Observed nested, crossing and disjoint ranges alone warn. Both narrower and wider deviations count; item prefixes and respondent overlap do not establish different constructs.
- Optional `permitted_values` supplies documented categories, shared or per item. Complete valid documentation resolves width warnings; an observed response outside a usable permitted set produces `resp_outside_permitted` failure. Valid entries in partial mappings are still checked.
- Optional `item_constructs` supplies explicit documented membership. A complete mapping identifying distinct constructs with different observed group ranges produces **`resp_scale_constructs`**, a separate failure name. It cannot be waived by an existing name-only `resp_scale_mixed` exception.
- `resp_outside_permitted` and `resp_scale_constructs` are errors under `triage`, `upload` and `legacy`. The three width-only checks remain warnings; `core` retains its R subset. Existing strict mode can still block warnings.
- Isolated outliers (strictly fewer than 15% of observed items) are classified before nesting. An HPQ-shaped table with twelve 1–5 items and `submiss` at 0–40 gets `item_scale_outlier`, naming that column and its range. Rare legitimate item formats also only warn. SDV-shaped 1–5/1–7/1–8/1–9 ranges remain nested-support warnings without construct evidence, the accepted tradeoff.

Validation:

- **123 validator tests pass** on both Python 3.12/pandas 2.2.3 and Python 3.14/pandas 3.0.5. These include **40 response-scale tests** covering majority reversals and ties, rare legitimate formats, the HPQ/SDV shapes, the 15% boundary, documented violations, partial/malformed evidence, profile behavior, API forwarding and numeric precision. A scoped-waiver regression proves that the old width-check name cannot waive the new documented-construct error.
- Nguyen regressions replay exact item marginal counts: Barthel 3,940 responses/10 items and MSPSS 4,728 responses/12 items. The fixture records source provenance and hashes, with no participant records. MSPSS's documented 1–7 categories come from its workbook codebook; Barthel's item-specific permitted sets are not inferred from observations.
- **65 uploader, 35 dictionary-staging, 4 item-text and 10 config-parser tests pass**, together with the R validator smoke and dictionary-union tests. The inherited literal-token regressions also pass; invalid text still blocks upload, and skipping width inference does not suppress a documented numeric violation.
- The text-only golden expectation is explicitly corrected: numeric failures remain, without a spurious numeric-range failure. The generic nonblocking-heuristic test now uses composite-looking names; response-width guarantees are covered in the dedicated suite.

These are software regression checks, not corpus false-positive estimates. HPQ/SDV tests use synthetic shapes; Nguyen marginal replays do not validate original participant pairing. Evidence inputs remain optional Python APIs through `context`, not automatic extraction or new CLI options. No data tables, conversion scripts or existing named waivers are changed. The PR is ready for review following the maintainer’s acceptance of this design.


sinew-07:
@ben-domingue I opened this as a draft because the implementation and regression tests are complete, but I’d especially value your feedback on the QC semantics—specifically whether valid item-specific nested response supports should remain a non-blocking note in the IRW workflow.

ben-domingue:
Thanks for this — and I want to lead with the fact that I think the **underlying idea is right and worth pursuing**. Separating "items legitimately have different response sets" from "two instruments got bundled into one table" is a real distinction that the current check collapses, and it has already cost us: `cao_2026_cdss`, `hoai_2026` and `chen_2026` all needed named `ALLOWED_FAILS` waivers for what turned out to be ordinary floor/ceiling non-use. Anything that cuts that manual adjudication down is welcome, and I'd like to see a version of this land.

But there's a complication I should raise now rather than after review, because it may be a deal-breaker for the framing as written.

### A scale is defined by the construct, not by the response structure

A math test can legitimately contain multiple-choice items scored 0/1 and constructed-response items scored 0-3 **in the same table**, because they measure the same thing. IRW's format handles this natively — `resp` is scored within item, and any polytomous model estimates per-item category thresholds. Mixed-format tests are ordinary, not defective.

The current check keys on response *width* as a proxy for construct identity. That proxy holds for Likert self-report, where one instrument means one uniform anchor set — which is why it caught genuine defects (`sdv` spanning 1-5/1-7/1-8/1-9 at once; `ywjhb9mw82` really mixing 6- and 7-point blocks). It does not hold for achievement data.

We already ship a counterexample. `mthimkhulu_2023_pirls_reading` is 1,894 learners × 15 items at `resp` 0-3, mixing constructed-response and multiple-choice items in one correct table. It clears triage only by accident: the `over` filter (`automated_finding/irw_triage_updated.py:715`) selects only items that *exceed* the modal range, so with 0-3 as the mode the narrower items read as ceiling non-use and pass.

Invert the proportions and it fails. Twenty MC items at 0/1 and five CR items at 0-3: modal range is 0-1, the five CR items exceed it, share is 20% ≥ 0.15 → `resp_scale_mixed` / `fail`, and the README then tells the submitter to "split into one table per scale" — splitting a single unidimensional test into two tables that can no longer be calibrated together.

So current behavior on mixed-format tests turns on whether MC or CR items are in the majority, a distinction with no psychometric content. And it bites hardest on dichotomously-scored achievement data, which the warehouse has far less of than self-report and is actively trying to grow.

### Two consequences for this PR

**1. The description and the implementation appear to have drifted.** The summary says the new logic compares *documented* response sets per item and fires when "the documentation is usable." The check on `main` reads no documentation at all — it's a pure observed per-item min/max profile (`irw_triage_updated.py:706-728`). Where do the documented sets come from? If the PR introduces that input, it's a substantially larger change than the four-file scope suggests; if it doesn't, the stated decision rule isn't the one being implemented.

**2. On your severity question,** I don't think `note`-vs-`fail` is quite the right axis. The deeper problem is that bare response-width heterogeneity shouldn't reach `fail` on its own. Width is evidence of bundled instruments only when it co-varies with something that actually marks a construct boundary — item-name prefixes (which `multi_scale*` already reads, immediately above at line 686), documented subscale membership, or disjoint respondent sets. Demoting the width-only signal to `warn`, and reserving `fail` for width-plus-a-construct-boundary, would keep `sdv` and `ywjhb9mw82` caught, let mixed-format tests through, and let us retire several waivers we're currently carrying.

That's the direction I'd be glad to see this go. Happy to look again once you've had a chance to consider it.

sinew-07:
> Thanks for this — and I want to lead with the fact that I think the **underlying idea is right and worth pursuing**. Separating "items legitimately have different response sets" from "two instruments got bundled into one table" is a real distinction that the current check collapses, and it has already cost us: `cao_2026_cdss`, `hoai_2026` and `chen_2026` all needed named `ALLOWED_FAILS` waivers for what turned out to be ordinary floor/ceiling non-use. Anything that cuts that manual adjudication down is welcome, and I'd like to see a version of this land.
> 
> But there's a complication I should raise now rather than after review, because it may be a deal-breaker for the framing as written.
> 
> ### A scale is defined by the construct, not by the response structure
> A math test can legitimately contain multiple-choice items scored 0/1 and constructed-response items scored 0-3 **in the same table**, because they measure the same thing. IRW's format handles this natively — `resp` is scored within item, and any polytomous model estimates per-item category thresholds. Mixed-format tests are ordinary, not defective.
> 
> The current check keys on response _width_ as a proxy for construct identity. That proxy holds for Likert self-report, where one instrument means one uniform anchor set — which is why it caught genuine defects (`sdv` spanning 1-5/1-7/1-8/1-9 at once; `ywjhb9mw82` really mixing 6- and 7-point blocks). It does not hold for achievement data.
> 
> We already ship a counterexample. `mthimkhulu_2023_pirls_reading` is 1,894 learners × 15 items at `resp` 0-3, mixing constructed-response and multiple-choice items in one correct table. It clears triage only by accident: the `over` filter (`automated_finding/irw_triage_updated.py:715`) selects only items that _exceed_ the modal range, so with 0-3 as the mode the narrower items read as ceiling non-use and pass.
> 
> Invert the proportions and it fails. Twenty MC items at 0/1 and five CR items at 0-3: modal range is 0-1, the five CR items exceed it, share is 20% ≥ 0.15 → `resp_scale_mixed` / `fail`, and the README then tells the submitter to "split into one table per scale" — splitting a single unidimensional test into two tables that can no longer be calibrated together.
> 
> So current behavior on mixed-format tests turns on whether MC or CR items are in the majority, a distinction with no psychometric content. And it bites hardest on dichotomously-scored achievement data, which the warehouse has far less of than self-report and is actively trying to grow.
> 
> ### Two consequences for this PR
> **1. The description and the implementation appear to have drifted.** The summary says the new logic compares _documented_ response sets per item and fires when "the documentation is usable." The check on `main` reads no documentation at all — it's a pure observed per-item min/max profile (`irw_triage_updated.py:706-728`). Where do the documented sets come from? If the PR introduces that input, it's a substantially larger change than the four-file scope suggests; if it doesn't, the stated decision rule isn't the one being implemented.
> 
> **2. On your severity question,** I don't think `note`-vs-`fail` is quite the right axis. The deeper problem is that bare response-width heterogeneity shouldn't reach `fail` on its own. Width is evidence of bundled instruments only when it co-varies with something that actually marks a construct boundary — item-name prefixes (which `multi_scale*` already reads, immediately above at line 686), documented subscale membership, or disjoint respondent sets. Demoting the width-only signal to `warn`, and reserving `fail` for width-plus-a-construct-boundary, would keep `sdv` and `ywjhb9mw82` caught, let mixed-format tests through, and let us retire several waivers we're currently carrying.
> 
> That's the direction I'd be glad to see this go. Happy to look again once you've had a chance to consider it.

Thanks, Ben — this makes sense. I agree that response-width differences alone should not trigger a failure. I’ll revise the rule so that width-only cases warn, while fail is reserved for cases with evidence of a construct boundary, and I’ll make sure the existing valid mixed-format cases still pass.

ben-domingue:
@sinew-07 i like the 'warn' idea. i think that would be valuable.

sinew-07:
> @sinew-07 i like the 'warn' idea. i think that would be valuable.

> Thanks, Ben — sounds good. I’ll implement the `warn` path and update the tests and documentation accordingly.

ben-domingue:
**Not a `data/` PR, so the new contract check does not apply — but the tests job does.**

CI landed today (#1821): every PR now runs `red_up` and `irw_validate`'s suites. This PR touches `automated_finding/irw_triage_updated.py`, and that file changed underneath you today — `run_qc` and `Check` moved into the new `irw_validate` package and are re-exported from their old names (#1821). **Nothing you import breaks**, but you will want to rebase; the conflict should be small.

Worth knowing for this PR specifically: severity is now a property of *(check, profile)* rather than of the check. Your `resp_scale_nested_support` work fits that directly — @ben-domingue's request that width-only signals be demoted to `warn` is now expressible without changing what triage reports, because `triage` and `upload` can disagree about the same check. `irw_validate/model.py` has the mechanism.

This is still marked **draft**, and @ben-domingue's review notes from 2026-08-30 are waiting on a push from you.
---

ben-domingue:
Read the whole diff and ran the suites locally (91 `irw_validate` tests pass). The restructuring into `irw_validate` is what I wanted, and the test coverage is genuinely thorough — the profile matrix, per-item vs shared codebooks, the float32/int64 precision cases. Two things to sort out before this comes out of draft.

**1. `resp_scale_mixed` now carries two meanings under one name, and that collides with existing waivers.**

The width-only path emits `resp_scale_mixed` as `warn`; the documented-constructs path emits the same name as `fail`; and the name is now in `GATE_ERRORS`. `severity_for` handles this correctly today — `status == "warn"` returns before the `GATE_ERRORS` lookup, so width-only findings do stay warnings.

The problem is downstream. `data/cao_2026_cdss.py:130` waives by name only:

```python
and (suffix, c.name) not in ALLOWED_FAILS
```

No status. So a table that currently carries a `resp_scale_mixed` waiver for the benign width reason would, as soon as anyone supplies `item_constructs` for it, silently suppress the real construct-boundary failure. Same pattern in `hoai_2026_blended_learning.py` and `soderberg_2024_esm_school.py`.

Please give the documented-evidence failure its own check name — `resp_scale_constructs` or similar. That keeps `GATE_ERRORS` unambiguous and stops name-keyed waivers from crossing the two. (As a bonus, once this lands those three waivers are dead entries and can be removed in a follow-up, which is the change proving its own point.)

**2. The two defects the old check was written for now only warn.** I reproduced both against this branch:

- The 2026-08-26 `sdv` build — items observed at 1-5, 1-7, 1-8 and 1-9, all sharing a floor — is `nested`, so it comes out as `resp_scale_nested_support` **warn**. It was a `fail`. Under `upload` it no longer blocks.
- The HPQ case: twelve items on 1-5 plus `submiss`, a missing-response count spanning 0-40, swept in as an item. That is also `nested` (the count column's range contains the item scale), so it too becomes `resp_scale_nested_support` and loses the `item_scale_outlier` diagnostic entirely — the branch that exists for exactly this shape. `item_scale_outlier` is still reachable for a *crossing* item (twenty items on 1-5 plus one on 2-7), just not for this one.

The first is the trade the PR is deliberately making, and I accept it — a nested support genuinely is not evidence of two constructs. The second I think is an accident of branch order: the nested test runs before the outlier-share test. Running the isolated-outlier test first would recover the swept-in-admin-column case without weakening the MC/CR fix at all. Worth a test pinning that case.

Minor, no action needed: the golden-fixture correction is right — the old check read raw `df["resp"]`, so text responses produced a spurious range failure. Note also that `test_heuristics_do_not_block_the_gate` was repointed from `resp_scale_mixed` to `composite_items*`; the original guarantee is still covered in `test_response_scales.py`, but that swap isn't mentioned in the description.

sinew-07:
Thanks, Ben — both points are addressed in 1be35c942ae4d5f76e7cd68fb68e658c350735b4, with current main incorporated and the conflict resolved.

- The documented-construct failure is now `resp_scale_constructs`, with its own `GATE_ERRORS` entry. A regression verifies that an old `resp_scale_mixed` waiver cannot suppress it.
- The isolated-outlier test now precedes nesting. The HPQ-shaped regression reports `item_scale_outlier` and names `submiss` (0–40); rare legitimate MC/CR formats still only warn. SDV-shaped nested ranges retain the warning-only tradeoff you accepted.

The validator suite passes 123 tests in both tested Python/pandas environments; GitHub's tests and contract checks also pass. HPQ/SDV are synthetic regression shapes, and the Nguyen fixtures remain marginal replays. The PR stays Draft.

ben-domingue:
Both asks are addressed — I re-read the diff and ran the shapes. `resp_scale_constructs` has its own name and `GATE_ERRORS` entry, the isolated-outlier test precedes nesting, and the HPQ shape names `submiss` (0–40) again. 123 tests also pass here on Python 3.12 / pandas 3.0.3.

Ratifying the design question so this can come out of draft: **advisory-only is the right landing point, and I'm accepting it deliberately, not by omission.**

To be explicit about what that means, since it's worth writing down: nothing in the repo supplies `permitted_values` or `item_constructs` — not `red_up`, not the CLI, not any `data/` script — so both new failures are reachable only from a hand-written Python call. On `upload` today, every response-scale finding is a warning. I ran sdv (1-5/1-7/1-8/1-9), the 6-vs-7-point block case, HPQ+`submiss`, and both MC/CR orientations against this branch: zero errors on all of them.

That's the correct trade. The check's false positives were costing real adjudication on correct tables; its true positives were two builds we caught anyway. And `model.py` already says heuristics cap at `warn` under `upload` with `GATE_ERRORS` growing one documented case at a time, so this is the house rule rather than an exception to it. If we later want width-only findings blocking, severity is a property of `(check, profile)` and that's a one-line change.

Your judgement call on the evidence sources — that item prefixes and disjoint respondent sets describe design as often as construct, so neither converts into membership — is the one I'd have made. Leave it as written.

**Please mark this ready for review; I'll merge it as-is.** Three follow-ups I'll file separately rather than grow this PR:

1. Wiring an evidence source, so the documented-failure paths have a caller.
2. Removing the now-dead `resp_scale_mixed` waivers. There are four, not the three I named: `cao_2026_cdss.py`, `hoai_2026_blended_learning.py`, `soderberg_2024_esm_school.py`, and `chen_2026_identity_construction.py:45`.
3. One narrowing to consider there: `resp_scale_constructs` requires documented constructs *and* differing observed ranges, so two documented constructs sharing 1-5 anchors still pass. Given explicit evidence the width conjunct may not be earning its place — but that's a question for when the path has a caller, not for this merge.

sinew-07:
Marked ready for review. Thanks for the review and for keeping the follow-ups separate.

ben-domingue:
Merging. Follow-ups filed as promised: #2152 (wire an evidence source, so the documented-failure paths have a caller — the width-conjunct question from point 3 is carried there, since it only becomes answerable once the path is callable) and #2153 (remove the four dead `resp_scale_mixed` waivers).

One thing #2152 records that is worth flagging here: item-text tables are keyed on `(item, resp)`, so a per-item permitted set already exists for 736 of 757 tables — shaped exactly like the `permitted_values` argument this PR added. The catch is that some of those sets look observed-derived rather than documented (`arzamoncunill_2023_epq_clinical__items` has `E1` at resp {1,2,4,5,6,7}, no 3), and feeding an observed set back in as documentation is circular. So the wiring needs a provenance gate rather than a blanket join. Not this PR's problem — but it means the evidence source you designed for is closer to hand than it looked.

Thanks for the two rounds of revision, and for holding the line on the evidence sources you declined to infer.

Issue #1645, opened 2026-08-18
Itemtext: verify the four earliest-uploaded tables, recoverable only from Redivis


ben-domingue:
`abouhashish_2025_chatgpt_attitudes`, `abukhalaf_2025_disaster_prep`, `addy_2021_sdq_ghana` and `agogue_2020_self_perceived_creativity` were uploaded in the first batch_001 pass and removed from disk before ever being committed, so they exist only on Redivis.

Every other uploaded table (27) was recovered from git history and swept with the current audit — including the new per-item resp-coverage check — and came back clean. These four are the only uploaded tables the current tooling has never checked.

**Action:** `irw::irw_itemtext(<table>)` for each, write to a scratch dir, run `audit_batch.R` over it. All four are `data_labels`, so risk is low.

ben-domingue:
Re-checked today: **the premise still holds.** These four have never been through the tooling.

`itemtables/batch_001/` contains exactly one item text file (`abdullah_2024_hpbbloat_stress__items.csv`), so `audit_batch.R` has never run on any of the four. They predate the batch system.

They *do* each have a `mapping_verification.csv` row, which could look like they've been checked. They haven't — all four read `data_labels` / `NOT_NEEDED`, dated 2026-08-17, i.e. retroactive provenance claims that the source file's own variable labels tie code to text. That is the strongest mapping basis when true, but it is an assertion made about these tables after the fact, not a check run against them.

So the work this issue asks for is unchanged: run `audit_batch.R` against live data, and confirm the `data_labels` claim by actually opening each source file. Small and bounded — four tables — and per #1680 still the best risk-per-hour on the item text list.

xingyi-zhang:
Done — all four verified. **PR #1737**, and the one real defect is filed separately as **#1738**.

Your premise held, and the retroactive rows were worse than "unchecked" — two of the four were wrong about *what kind* of mapping they described.

`audit_batch.R` returns **4 PASS**. That was never going to be the interesting half: the defect below is invisible to it, because the resp *set* matches live data exactly and the error is in which anchor sits on which value inside that set.

## 🔴 addy_2021_sdq_ghana — SDQ_7 anchors reversed (#1738)

The source `.sav` assigns one value-label set to every SDQ item (`0=not true, 1=somewhat true, 2=certainly true`). The curation ships `SDQ_7` flipped: `0=certainly true`, `2=not true`. So a respondent who answered `0` is recorded as having said the opposite.

[...]

PR #1737, opened 2026-08-30
Verify the four earliest-uploaded itemtext tables (#1645); fix addy SDQ_7 anchors

xingyi-zhang:
Closes the verification work in #1645.

Those four batch_001 tables were uploaded and then removed from disk before being committed, so they existed only on Redivis and had never been through the tooling. Recovered with `irw::irw_itemtext()` into `audit_staging/issue1645/` (gitignored), audited against live data, and — per #1645's comment, the part that had never actually been done — checked against their real source files.

`audit_batch.R` returns **4 PASS**. The one real defect is invisible to it, because the resp *set* matches live data exactly and the error is in which anchor sits on which value inside that set.

## 🔴 `addy_2021_sdq_ghana` — SDQ_7 anchors reversed

The source `.sav` assigns one value-label set to every SDQ item — `0=not true, 1=somewhat true, 2=certainly true`. The published curation ships `SDQ_7` flipped:

| item | resp 0 | resp 1 | resp 2 |
|---|---|---|---|
| `SDQ_1`–`SDQ_6`, `SDQ_8`–`SDQ_10` | not true | somewhat true | certainly true |
| **`SDQ_7`** (published) | **certainly true** | somewhat true | **not true** |
| **`SDQ_7`** (corrected here) | not true | somewhat true | certainly true |

`SDQ_7` is "i usually do as i am told", one of the SDQ's five canonical reverse-scored items, so the flip looks like an assumption that the stored data is pre-reverse-scored. It isn't: `r(SDQ_7, prosocial mean of SDQ_1/4/9) = +0.156` — positive, which is what raw storage predicts — and the source's own labels read `0=not true` for `SDQ_7` like every other item.

Corrected file is in `fixes/` with a `diffs_vs_published/` writeup. **2 rows changed, nothing else touched.** Not uploaded — `upload.py` replaces a table wholesale, so that needs your explicit approval.

The item text itself is *perfect*: all 10 values byte-identical to the source variable labels, source typos preserved ("alot", "oten lose my temper", "fidgiting"). Do not tidy those.

## The four mapping_verification rows

All four carried identical boilerplate evidence dated 2026-08-17 — "Item code is tied to its text by the source data file's own variable labels… Numeric verification would add nothing." Replaced with what was actually checked:

| table | was | now |
|---|---|---|
| `addy_2021_sdq_ghana` | data_labels / NOT_NEEDED | data_labels / **VERIFIED** |
| `agogue_2020_self_perceived_creativity` | data_labels / NOT_NEEDED | data_labels / **VERIFIED** |
| `abouhashish_2025_chatgpt_attitudes` | data_labels / NOT_NEEDED | **paper_order** / VERIFIED |
| `abukhalaf_2025_disaster_prep` | data_labels / NOT_NEEDED | **reconstructed** / **PARTIAL** |

The two reclassifications are the substantive part:

- **`abouhashish` was not `data_labels` at all.** Its codes are `item_01..item_40` — positional indices assigned by the processing script, not source column names. That's the pattern the skill flags as highest-risk and where every mapping defect found in review has come from, and it had been recorded as *exempt from verification*. It turns out to be correct — I diffed all 40 shipped texts against the source `.xlsx` header at each position and the alignment holds — but that check had never been run. Two deviations disclosed rather than fixed: the headers are bilingual English/Arabic and only the English shipped, and whitespace was normalised in two items.
- **`abukhalaf` is `PARTIAL`, not verified.** Codes are script-assigned mnemonics (`prep_kit`, `prep_evac_plan`, `prep_comm_plan`). With three items drawn from three adjacent headers, no route distinguishes them from each other — the tie rests on reading the processing script, not on evidence in the data. Also worth a website caveat: per the paper, `resp` is a **proportion**, not an ordinal level ("answers were translated into percentages… Strongly Disagree (0.0%)… Strongly Agree (100%)").

`agogue`'s partially-blank `option_text` (endpoints only) is **correct, not a gap** — the `.sav` carries no value labels and the paper states only the endpoints.

## Notes for review

- **A false signal to be aware of:** the first audit pass reported 4 WARNs including "100% of rows have blank `option_text`". All were artifacts of how these files were exported here (`irw_itemtext()` → `write.csv`), not properties of the curation; `normalize_nulls.R` cleared them all. `audit_batch.R`'s blank-`option_text` check is sensitive to whether absence is stored as `""` or the `NA` token.
- `mapping_verification.csv` is CRLF in this repo (Python `csv.writer` default) while `write.csv` emits LF, so a naive read/write round-trip reformats all 126 lines. Restored deliberately — the diff here is **4 changed rows and nothing else**.
- Two response-data items surfaced that are out of scope for itemtext and not addressed here: addy's `SDQ_6`/`SDQ_10` contain no zeros at all and correlate with nothing, and the source file's `SDQ*_CAT` family is offset one column from `SDQ_n` (`SDQ(n+1)_CAT = (SDQ_n+2) mod 3` at 100%). Both are recorded in `diffs_vs_published/addy_2021_sdq_ghana.md`.


xingyi-zhang:
Added a second commit (`5d83d2a`) carrying follow-up 3 from the batch report: the `abukhalaf_2025_disaster_prep` caveat for `itemtext_issues.qmd`.

That file lives in `datapages/irw`, where I have `pull` but not `push`, so it cannot land from here (access requested on #1645). Rather than leave it depending on a local site clone, it is stored as `git format-patch` output at `itemtext/fixes/itemtext_issues_abukhalaf_2025_disaster_prep.patch` — `git am` it from the root of a `datapages/irw` checkout and it brings its own message and authorship. Verified to apply cleanly against site HEAD `befb1e9`; 7 lines, div balance 171/171.

No change to the four files already under review here.

Issue #1751, opened 2026-08-31
itemtext: test the Dataverse API route against the WAF-blocked tables, starting with gilbert_meta_109/_111


xingyi-zhang:
Assigned by @ben-domingue on #1696 (which is now closed, so filing it here to work under). Parent roadmap: #1709, sub-action 7.1.

## The question

**Does the Harvard Dataverse API reach sources whose web pages are WAF-blocked, and if so how many `BLOCKED` tables stop being blocked?**

#1615 established that it does for one table. From that thread:

> The Harvard Dataverse **API** is not WAF-blocked even though the web page is, so `api/access/datafile/...` pulled both the original `.dta` and `Replication_do_file.do`.

That single access unblocked `gilbert_meta_35` and, in the process, showed its **published item text was wrong on 7 of its 8 items** — the scale was `a,c,d,e,f,g,i,j`, not `a`–`h`. So the API route has already paid for itself once.

`gilbert_meta_109` and `gilbert_meta_111` are the same family in the same repository, and neither has any item text yet.

## Why `BLOCKED` is the wrong word for these

`BLOCKED` records that nobody could reach the source, not that the source lacks item text. Both audit rows say so explicitly:

| table | classification | reasoning |
|---|---|---|
| `gilbert_meta_109` | BLOCKED | "Dataverse page returned no fetchable content, consistent with the known AWS WAF bot-challenge." |
| `gilbert_meta_111` | BLOCKED | "Page returned no content (AWS WAF bot-challenge), matching sibling table `gilbert_meta_109`'s confirmed BLOCKED status from the same record." |

Neither says the text isn't published. `_111`'s verdict is weaker still — inferred from `_109` rather than checked independently. That is the trap SKILL.md Step 3 names: "a verdict of 'text not published' that was reached without opening the tables is not final."

## Scope

From `itemtext/availability_audit_full.csv` (2,110 rows: 1,348 AVAILABLE, 217 BLOCKED, 390 UNAVAILABLE, 155 UNAVAILABLE-copyrighted):

| set | n |
|---|---|
| BLOCKED total | 217 |
| …citing a Dataverse/Harvard source | 53 |
| …citing WAF / bot-challenge / "no fetchable content" | 46 |
| …**both** (the set this route directly addresses) | **36** |
| …either | 63 |

[...]

Issue #1696, opened 2026-08-28
itemtext onboarding: extract and spot-check mohammed_2021_job_satisfaction


ben-domingue:
## Onboarding: run the item text skill on one table, then check its work

Welcome! This is a scoped first task on the IRW item text pipeline. The goal is not the
table — it's for you to see how the extraction process works end to end, and to come away
with your own view of how far the automation can be trusted.

**Start here: [Item Text Extraction Runbook](https://claude.ai/code/artifact/7c598820-03ce-418b-b2a8-720352448c08)**

Read it before you start. Sections 1 (what you're producing), 3 (the per-table pipeline)
and 4 (the gates and what each cannot see) are the ones you need for this task; the rest
is context for later.

---

### 1. Extract

Working directory is `irw/src/itemtext/`. Ask Claude Code to run the item text skill on:

```
mohammed_2021_job_satisfaction
```

The skill is `irw-auto-itemtext` — "generate item text for mohammed_2021_job_satisfaction"
will trigger it. Let it work through SKILL.md Steps 2–6 rather than steering it; the point
of this exercise is to see what the process produces on its own.

**Expected output:** one file, `<table>__items.csv`, plus a `provenance.csv` row.
**Please stop there — do not upload anything.** Uploading to Redivis is a separate,
explicit step that Ben triggers.

Two scope notes:
- `mohammed_2021_patient_safety_culture` is a sibling table built from the same source
  file (42 HSOPSC items). **Leave it alone** for this pass — it's a separate queue row.
- Anything named `enem*` is permanently out of scope; Ben handles those separately.

---

### 2. Why this table

It was picked to be fully checkable by hand in about twenty minutes, which is what makes
it useful for building — or withholding — trust:

- **Small and complete.** 5 items, a 1–5 response scale, 411 respondents, 2,055 rows, no
  missingness quirks. Every item uses all five levels.
- **The source is open.** PLOS ONE, CC BY: Mohammed et al. (2021),
  [10.1371/journal.pone.0245966](https://doi.org/10.1371/journal.pone.0245966). The S1
  File is an SPSS `.sav` anyone can download.
- **The mapping is authoritative at the source**, so there's a right answer you can look
  up rather than infer. This is the `data_labels` case in the runbook's section 5.
- **The item codes are opaque on purpose.** They are `jobsat`, `jobsats`, `jobsatss`,
  `jobsatsss`, `jobsatssss` — five codes distinguished only by trailing `s` characters.
[...]

PR #1752, opened 2026-08-31
Reach WAF-blocked Dataverse sources via the API; re-audit gilbert_meta_109/110/111 (#1751)

xingyi-zhang:
Answers #1751. **The API route works — uniformly — but access and availability came apart on the first record tried, so the two tables assigned still end up without item text.** Both halves matter.

## The route works

| | result |
|---|---|
| `dataset.xhtml?persistentId=…` (web page) | **13/13 → HTTP 202, 0 bytes** |
| `api/datasets/:persistentId` | **12/12 records with a resolvable DOI → HTTP 200** |
| records holding ≥1 ingested file with an `originalFileFormat` | **11 of 12** |

The one API failure is a dictionary gap, not a WAF one — `karajko2025_*` has no `persistentId`-style URL, so no DOI to resolve.

`?format=original` is the step that matters. Dataverse ingests `.dta`/`.sav` into a tab-separated archival copy **with the variable labels stripped**, and the labels are the whole point. Recipe is in `itemtext/dataverse_api_route.md`.

**So a `BLOCKED` verdict reached by fetching a Dataverse page records the access method, not a property of the source.**

## …and it does not guarantee item text

`doi:10.7910/DVN/LUKGIA` (Mosher & Kim 2025 — `gilbert_meta_109`/`_110`/`_111`) is *fully* reachable: 6 files, the original `.dta` with labels, the replication do-file, the readme. The item wording still is not published in any of them:

- `.dta` variable labels: `recall q1`…`recall q5`, `mid transfer q1`…`q13`. These **pin the mapping** but are not the administered wording — the `bitew_2020_osss3` case.
- `.dta` value labels: none on the item variables.
- `structured_supplements_public.do`: **zero** `label variable` statements — it's an analysis do-file.
- `readme.docx`: describes the instruments ("a 5-item recall assessment designed to test students' ability to recall basic facts from the read aloud story", scored 1/0) and explicitly defers to the paper.
- The paper's **openly readable** accepted manuscript (Harvard DASH): "a 5-item multiple choice assessment (with 4 options per item)", alpha 0.63; near/mid-transfer each "13 multiple-choice items with four options per item". Reports the psychometrics, does not reproduce the items.

Also worth noting: `resp` is a 0/1 correctness score, not an option choice, so `option_text` would be behaviour-scored blank even if the wording turned up.

`gilbert_meta_109` and `_111` therefore move **BLOCKED → UNAVAILABLE** — a settled answer rather than an unexamined block, which is the "couldn't do this one" outcome I'd asked to see.

## Two audit-hygiene problems this exposed

**1. Siblings from one source file were classified inconsistently.** `data/gilbert_109through111.R` builds all three tables from one `.dta`, yet:

| table | was | assessed from |
|---|---|---|
| `_109` | BLOCKED | the web page |
| `_110` | UNAVAILABLE | `data/gilbert_109through111.R` + the raw `.dta` |
| `_111` | BLOCKED | **inferred from `_109`** |

One sibling had already been assessed from the raw file and answered the question, while the other two were marked blocked from a page fetch. Worth checking for the same pattern elsewhere.

**2. A verdict can be right for a false reason.** `_110`'s note read "No source paper for this structured supplements reading-intervention RCT could be located via search." The paper is cited in the dictionary (`10.1080/10888438.2024.2368145`) and openly readable on Harvard DASH. `UNAVAILABLE` still holds, but on that reasoning nobody would ever have retried it. Corrected here; classification unchanged.

## A count to settle

#1696 cites "62 of the 217 BLOCKED are Dataverse WAF bot-challenges". By the audit's own text: **36** rows cite both a Dataverse/Harvard source and a WAF/bot-challenge, 53 cite Dataverse, 46 cite WAF, and **63** cite either. The 62 looks like the union rather than the intersection. The tight target class is 36, of which 13 have a resolvable DOI today — **the other 23 need DOIs before they can be re-audited this way**, which is the real blocker on scaling this.

## Note on the diff

`availability_audit_full.csv` was edited **line-by-line, not round-tripped**. Its own quoting is inconsistent (645 rows unquoted, ~1465 quoted, header unquoted), so no single writer rule reproduces it — `read.csv` → `write.csv` turned a 3-row change into a 2111-row diff, and Python's `QUOTE_MINIMAL` into 687. The diff here is **3 rows**. Same class of hazard as the CRLF one on `mapping_verification.csv`, different axis.

## Not in scope

No item text shipped — there is none to ship for these tables. Nothing uploaded.


xingyi-zhang:
Added `4b06a00` after a good challenge from @xingyi-zhang: a human with a browser and institutional access might get further than I did.

**Partly right, and it caught a real gap.** My appendix check had been reading DASH's auto-extracted **TEXT** bundle, which truncates at char 99,564 of 99,652 — exactly where `Supplementary Appendices` begins. It carried the appendix headings and none of their content, so "the appendices are unexaminable" was an artifact of fetching the wrong bitstream.

Fetched the `ORIGINAL` bundle instead. The 71-page PDF carries all six appendices, and **S-A.1–S-A.6 are balance tables, a Benjamini-Hochberg correction and an SEM sensitivity check** — zero occurrences of `item`, `question`, `passage` or `option`. Verdict stands, now on the full document rather than a truncated extract.

**Where human access definitively does not help:** the Dataverse record is exhausted. `license: CC BY-NC-SA 4.0`, `restricted=False` and no embargo on all six files — the API already had the complete public set, so a browser or institutional login adds nothing there. `dataFile.restricted` and `latestVersion.license` in the record JSON tell you which situation you are in for one request, and that check is now in the write-up.

**Where it still might:** two off-Dataverse routes, recorded in the audit rows as *not yet tried* so the verdict is scoped to what was actually checked — the publisher's supplementary data at tandfonline/Ingenta (403 to automated access, and may differ from the accepted manuscript), and emailing READS Lab, whose readme names `ethan_scherer@gse.harvard.edu` for data requests. `processing_notes/Licensing.txt` has the template.

Both traps are now in `dataverse_api_route.md`, since the truncated-TEXT-bundle one would bite anyone auditing a DASH-hosted paper.

Issue #1744, opened 2026-08-30
auto-tag: 37% of tables yield no tags because the source can't be fetched — 11 of the 22 look fixable


ben-domingue:
In the 2.2 scoring run (#1721), **22 of 60 sampled tables produced no tags at
all** — 37%. Every one of the 60 had a dictionary entry with a URL or DOI, so
this is not a metadata gap. The source text could not be retrieved.

| reason | n | fixable? |
|---|---|---|
| `fetch_failed` | 11 | **probably** |
| `paywall` | 9 | no |
| `no_source` | 2 | no |

**This caps yield before accuracy enters the picture.** Even a perfect tagger
reaches only ~63% of tables. Improving `sample` from 42% lifts one column;
recovering the 11 fetch failures lifts every column at once, for the same
tables.

### The 11

    aspirations_sonmez_2022                       w1
    gilbert_meta_40                               w1
    cfc_balaji_2019                               w1
    enem_2022_1mil_cn                             w1
    pisa2009_math                                 w1
    ecuador_2011_safety_crime                     w2
    gahps_korner_2021                             w2
    jeilani_2024_personal_growth                  w2
    jeilani_2024_self_efficacy                    w2
    c19prc_uk_mcbride_2021_socialdistance         w4
    c19prc_uk_mcbride_2021_polarization_brexit    w4

Two patterns are visible without any further work:

1. **Assessment programmes, not papers.** `pisa2009_math`, `enem_2022_1mil_cn`,
   `ecuador_2011_safety_crime` and the `c19prc_uk_*` tables point at data
   portals and programme sites rather than a journal article. `fetch_source.py`
   is built to resolve a DOI to a paper; a portal landing page is a different
   shape of problem and may need a different strategy (or an explicit "no
   paper" verdict, so it stops looking like a transient failure).
2. **Repeated sources.** Both `jeilani_2024_*` tables failed, and both
   `c19prc_uk_*` ones did. A single unreachable source can take several tables
   with it, so the 11 failures are fewer than 11 distinct causes.

### Scope

1. Re-run `fetch_source.py` on the 11 and record the actual failure for each —
   timeout, non-200, redirect loop, robots block, JS-only page, PDF that yields
   no text. Right now "fetch_failed" is one bucket hiding several causes.
2. Fix what is mechanical (retries, redirects, PDF extraction, user-agent).
3. For the portal-shaped ones, decide whether a distinct verdict is warranted
   so they are not retried forever as if transient.
[...]

Issue #1721, opened 2026-08-29
[2.2] Score the auto-tagger per column against the human gold set


ben-domingue:
Sub-action **2.2** of #1704. **This is the current first priority.**

The tag gap has two possible causes and we cannot presently tell them apart:

1. **The tagger is good and the write path is the bottleneck** — then the fix is 6.1 (#1708) plus a bulk write, and the gap closes in weeks.
2. **The tagger is not good enough yet** — then the fix is vocabulary and prompt work, and 6.1 is premature.

Every other sub-action under #1704 is conditional on the answer. This one distinguishes them using data that already exists, requires no new labelling, and **does not depend on 6.1** — the tagger can be scored without ever writing to the Sheet.

### Scope: the four closed-vocabulary columns first
`item format` (5 values), `age range` (6), `measurement tool` (6), `child age` (10) can be scored today with no prep. `sample` (52) and `construct type` (66) must wait for **2.1**.

### Method
- Sample ~60 already-tagged tables, stratified across warehouses so w3 and w4 are represented
- Run the `irw-auto-tag` skill on them blind
- Compare per column against the human tags; report accuracy plus the confusion pattern for each

### Done when
Four accuracy numbers exist, and there is a defensible bar for **2.3**.

### Why it matters
If the four clean columns score well, half the tag schema can ship without waiting on 6.1, the remaining columns, or anything else.

---
Parent: #1704 &nbsp;·&nbsp; [full context](https://claude.ai/code/artifact/a322f671-38af-48c1-b4c8-1e80cbf8ec51#i2)

ben-domingue:
@jdtrinh & @stacycremer can you add details of your original tagging test here so that we can incorporate these into planned next steps?

ben-domingue:
To make that easier to answer, here are the five specifics that would let us slot your test straight into the plan above:

1. **Which of the 8 tag columns did you cover?** The four closed-vocabulary ones (`item format` / 5 values, `age range` / 6, `measurement tool` / 6, `child age` / 10) are a very different problem from `sample` (52 distinct values in the sheet today) and `construct type` (66) — so results on one group don't transfer to the other.
2. **What did you score against** — the existing human tags already in the sheet, or your own independent read of each table?
3. **How many tables, and were they spread across warehouses?** w3 (27% tagged) and w4 (35%) are the undertagged ones and may differ in character from w1 (99.6%).
[...]

Issue #1702, opened 2026-08-29
Year 3 infrastructure: where to invest (17 numbered proposals)


ben-domingue:
**Full write-up (numbered, linkable):** https://claude.ai/code/artifact/a322f671-38af-48c1-b4c8-1e80cbf8ec51

A fresh look across all four repos (`src`, `irw_site` incl. all 7 branches, `Rpkg`, `Python-pkg`), the BRM and CEJEME papers, and the Y2 APR, asking where Year 3 infrastructure effort should go.

### The read

In 16 months the IRW went 892 → 3,650 tables. The growth has been in **acquisition** — `automated_finding/` is now genuinely good software. Almost every layer *downstream* of it is still sized for the ~900-table warehouse, and the gap is now user-visible (`itemresponsewarehouse/Rpkg#120`: `irw_filter()` returns tables `irw_fetch()` can't retrieve).

| | |
|---|---|
| Tables live in Redivis | ~3,650 |
| …with a row in `metadata.csv` | 89% — **403 missing**; warehouses 5 and 6 have *none* |
| …with tags | **61.7%** (w3: 27%, w4: 35%); 184 orphan tag rows |
| …with item text | **12.7%** — against 805 known to have it |
| `datastandard.md` checks that are executable | **2 of 10** |
| Automated tests in `src/` or `data/` | **0** |
| Ways to fetch a pinned IRW version | **0** |

---

### Corrected 2026-08-31

Measured two days after this was written, against `metadata/metadata.csv`, `metadata/tags.csv` and `metadata/itemtext_metadata.csv` (regenerated 2026-08-31), joined **case-insensitively** — a case-sensitive join reports 502 orphan tag rows instead of 194, because 308 tag rows differ from their metadata row only by case. The table above is left as written; the direction of travel is the finding.

| | as written 2026-08-29 | measured 2026-08-31 |
|---|---|---|
| Tables live in Redivis | ~3,650 | **4,134** |
| …with a row in `metadata.csv` | 89% — 403 missing; w5/w6 have *none* | **100% — 0 missing, in both directions** (w5: 206, w6: 1) |
| …with tags | 61.7% (2,251) | **55.3% (2,286)** — *down 6.4 points* |
| …with item text | 12.7% | **13.5%** (558 of 4,134) — see denominator note |
| orphan tag rows | 184 | 194, and **0 published** as of #1766 |

**Two of these change the argument, not just the digits.**

[...]

PR #1756, opened 2026-08-31
Reach open-access sources via OpenAlex, and stop caching blocker pages (#1744)

ben-domingue:
Closes the fixable half of #1744, and corrects one of its premises along the way.

### What #1744 said

37% of tables in the 2.2 scoring run (#1721) produced no tags at all — 11 fetch failures, 9 "paywalls", 2 with no usable source — and reachability, not accuracy, is the ceiling on all seven tag columns. The follow-up comment established via OpenAlex that 6 of the 9 "paywalled" papers have a legally open version, so the fixable pool was ~17 of 22.

### What this changes

`fetch_source.py` did a single `GET` against `doi.org` and cached whatever came back. Publisher landing pages routinely refuse an unauthenticated GET even when the article is fully open, so the tagger saw a blocked fetch and wrote `Notes = "cannot fully access due to paywall"`.

Now, given a DOI, it asks OpenAlex for open locations and tries them (PDFs first, extracted with `pypdf`) before falling back to the DOI and then the dictionary URL. It reports OpenAlex's `oa_status`, which turns "paywalled" into an asserted fact rather than an inference from failure. `SKILL.md` Step 3 acts on the distinction:

- `UNREACHABLE oa_status=closed` → a real paywall, stage the paywall note
- `UNREACHABLE oa_status=gold|green|hybrid|bronze|diamond` → an open copy exists and a WAF/captcha/JS page blocked us. **Not a paywall**, and fixable.

### The part that needed a second pass

A first version of this scored **14 of 22 reachable** by checking byte count. Five of those were blocker pages — including `abdullah_2024_hbbloat_pbc`, the table #1744 cites as the clearest case of a false paywall, which returned a 21kB **Google reCAPTCHA challenge**. OSF's JavaScript shell is a constant 4207 bytes and also passed. Byte count is not a content check.

Responses are now stripped to visible prose and rejected on blocker markers or under 1500 characters, with a per-candidate reason (`blocker:recaptcha`, `too_short:3_chars_visible`, `http_403`).

### Result

**0 → 9 of 22** reachable with verified prose. Artifacts and the reproduction commands are in `tags/scoring/reachability_2026-08-31.txt`.

The 13 that still fail, and why:

| Group | Tables | Reason |
|---|---|---|
| OECD/PISA | `pisa2022_read`, `pisa2009_math` | `http_403`, `oa_status=closed` — genuinely closed, and the real source is PISA technical documentation, not an article |
| Harvard Dataverse | `gilbert_meta_40`, `cfc_balaji_2019` | `http_202` WAF interstitial — same shape as #1751, which is already testing the Dataverse API route |
| Figshare | `jeilani_2024_personal_growth`, `jeilani_2024_self_efficacy` | `http_202`, same shape |
| OSF landing pages | 4 tables | JavaScript-only; the OSF API would reach these |
| No dictionary link | `spain_2023_identity_authenticity`, `spain_2023_identity_territorial` | Correctly a `no working link` case — unchanged by this work |

So the remaining gap is mostly **API routes for three known hosts** (Dataverse, Figshare, OSF), not access rights. #1744's conclusion that institutional access is off the critical path holds — only the two OECD tables are genuinely closed.

### Found along the way, filed separately

`oxfordcovid_xue_2024_brs` now fetches successfully and returns *"A Stan tutorial on Bayesian IRTree models"* — the dictionary's `doi` column names a different work than its `reference` column. A successful fetch can still feed the tagger the wrong paper, which no amount of reachability work catches.

Refs #1702 (item 2).

### DIFF
--- a/tags/.claude/skills/irw-auto-tag/SKILL.md
+++ b/tags/.claude/skills/irw-auto-tag/SKILL.md
@@ -103,18 +103,33 @@ proceed.
 ## Step 3 — Fetch the source text
 
 ```bash
-python3 scripts/fetch_source.py table_a --doi 10.1037/a0022874   # prefer DOI when present
-python3 scripts/fetch_source.py table_a --url https://...        # else use the dictionary URL
+python3 scripts/fetch_source.py table_a --doi 10.1037/a0022874 --url https://...
 ```
-This resolves `https://doi.org/{doi}` (or fetches the URL directly),
-caches the raw text under `.cache/` (gitignored — not committed), and
-reports HTTP status / final URL / byte count. It does **not** decide
-whether the content is paywalled — read the cached file yourself (`Read
-.cache/{table}.txt`) and judge that. **If it's paywalled, login-gated, or
-otherwise unreachable/unusable**, stop for that table and stage a row with
-every field blank except `table`, `Rater`, and
-`Notes = "cannot fully access due to paywall"` — don't guess content from
-the Description/Reference text alone.
+Pass **both** when the dictionary has both — they are tried in order, not
+as alternatives. Given a DOI the script asks OpenAlex where an open copy
+lives and tries those locations (PDFs first, extracted to text) before
+falling back to resolving the DOI itself, then the URL. It caches under
+`.cache/` (gitignored — not committed).
+
+It reports `oa_status` from OpenAlex, and **that is what you judge a
+paywall on — not the fetch failing.** Those are different findings:
+
+| Outcome | What it means | What to do |
+|---|---|---|
+| `OK ... content=N_chars_visible` | Got prose that is not a blocker page | Read `.cache/{table}.txt` and continue to Step 4 |
+| `UNREACHABLE oa_status=closed` | No open copy exists | Stage `Notes = "cannot fully access due to paywall"` |
+| `UNREACHABLE oa_status=gold\|green\|hybrid\|bronze\|diamond` | An open copy **exists** and something else blocked us — a WAF, a captcha, a JS-only page, a dead link | **Not a paywall.** Stage `Notes = "no working link"`, and it is worth reporting: these are fixable |
+
+The script rejects blocker pages rather than caching them, and prints the
+reason per candidate (`blocker:recaptcha`, `too_short:3_chars_visible`,
+`http_403`). This matters because raw byte count cannot tell an article
+from a challenge page: the reCAPTCHA interstitial served for a fully
+open-access PeerJ paper was 21kB, and OSF's JavaScript shell is a
+consistent 4.2kB. Both used to cache and read as "the source".
+
+Still read the cached file and judge it yourself — the script screens out
+blockers, it does not confirm the text is the *right* paper. A dictionary
+`doi` can point at a different article than the `reference` field names.
 
 Cached text persists across runs specifically so a paywalled or
 rate-limited source isn't re-hit on every retry.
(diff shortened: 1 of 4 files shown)


ben-domingue:
## Follow-up: the first number was rate-sensitive, so PMC now goes through an API

Smoke-testing this branch after it was pushed, `heise_2024_mental_falsesign` **failed** — the same table that had fetched 121k characters of article text an hour earlier. NCBI serves a reCAPTCHA to repeated unauthenticated callers, so scraping `pmc.ncbi.nlm.nih.gov` degrades as you use it.

That matters beyond one table: **a reachability figure measured by scraping decays on re-run**, which makes it useless as a threshold for 2.3 — you would set a bar against a number that quietly gets worse.

PMC-hosted articles now resolve through the **Europe PMC REST API** (DOI → `pmcid` → `fullTextXML`): a real API returning JATS XML, no challenge for repeat callers. Scraping stays as the fallback for records Europe PMC does not carry — `dwyer_2025_genomics_pb` is `inEPMC` but not open access, so its `fullTextXML` 404s and the PMC page still serves it.

Also added a 1s courtesy delay between attempts, and silenced bs4's `XMLParsedAsHTMLWarning`, which was writing noise into the stderr the skill has to read.

### Updated result

**0 → 10 of 22** (was 9), and the PMC ones are now stable across runs rather than degrading.

The headline case: `abdullah_2024_hbbloat_pbc` — cited in #1744 as the clearest false paywall, and the table that defeated *both* the original code and the first version of this fix — now returns **110,275 characters** of full text.

The 12 remaining failures are unchanged and are all host-routing, not access rights: OECD 403s (genuinely closed), Dataverse/Figshare `http_202` WAF interstitials (#1751), OSF JS-only pages, and two tables with no dictionary link at all.

One honest caveat now recorded in the artifact: `ecuador_2011_safety_crime` clears the 1500-char floor at 4419 characters but is a portal navigation menu, not a description of the instrument. It is a weak pass, and a stricter content check would drop it to 9.

ben-domingue:
## Third pass: verifying the text is the work we asked for

The `oxfordcovid_*` dictionary entries were [fixed at source](https://github.com/ben-domingue/irw/pull/1756) — all 17 now carry `10.5334/jopd.56`, matching their `reference` field, instead of the unrelated IRTree tutorial. Re-fetching with the corrected DOI turned up a further failure mode in this script.

The JOPD galley URL for that DOI redirects to the journal's **article listing page**: 12,722 characters of genuine prose, no blocker markers, comfortably over the length floor. It passed every check on this branch — and it is not the paper. The tagger would have read a page of unrelated article titles and extracted sample characteristics from it.

Fetched text must now contain either the requested **DOI** or a prefix of the work's **title** from OpenAlex; otherwise the candidate is rejected as `wrong_work` and the next location is tried. `--url`-only fetches (the data-portal case) pass unchecked, since there is nothing to check against.

With the corrected DOI, `oxfordcovid_xue_2024_brs` now reaches the actual ARC data paper — 41,860 characters that mention the Brief Resilience Scale and adolescent participants, which is exactly the context the tagger needs and never previously received.

### Scope of the check, stated precisely

It verifies *the text matches the DOI it was asked for*. It **cannot** catch a dictionary entry naming the **wrong** DOI — with the old `oxfordcovid` entry the fetch would have been confidently, verifiably correct about the IRTree paper. Those are two different defects and only the first is fixable here; the second is a dictionary-integrity problem.

Given the dictionary supplies `doi`, `url` and `reference` independently, a `doi`-vs-`reference` consistency sweep across the whole dictionary is probably worth doing as its own piece of work — `oxfordcovid` was 17 tables found by accident.

### Result

**10 of 22**, unchanged — the identity check costs no true positives while removing a false one.

Issue #1743, opened 2026-08-30
auto-tag: vocab.md never says a range crossing 18 is `Mixed` — 9 of 10 age_range errors


ben-domingue:
Nine of the ten `age_range` errors in the 2.2 scoring run (#1721) are the same
single mistake:

    gold `Mixed`  ->  predicted `Adult (18+)`     9x
    gold `Adult (18+)`  ->  predicted `Mixed`     1x

`age_range` scored **73.0%** accuracy. Essentially all of the gap is this one
confusion, so it is not a distribution of errors to grind down — it is a missing
rule.

### The rule

Human raters code a sample spanning e.g. 15–82 (mean 42) as `Mixed`, because it
includes minors. The tagger reads "adults were recruited", sees an adult mean
age, and commits to `Adult (18+)`.

`references/vocab.md` currently lists the five `Age Range` values with no
guidance on choosing between them. It never says that a reported *range*
crossing 18 makes the answer `Mixed` regardless of the mean or of how the
authors describe their sample.

### Fix

Add that rule to the `Age Range` section of
`tags/.claude/skills/irw-auto-tag/references/vocab.md`. Something close to:

> Choose `Mixed` whenever the reported age range spans 18, even if the mean is
> adult and the paper describes the sample as adults. `Adult (18+)` means no
> participant is under 18. When the paper gives only a mean and no range, and
> the population is clearly adult, `Adult (18+)` is right.

Then re-run the 2.2 scoring for this column against the same 60-table sample
(`tags/scoring/`, `seed=1723`) and confirm it moves. If it does, `age_range`
graduates from "hold" to "ship" in 2.3 (#1722).

### Why it is worth doing first

It is one sentence, it is testable against a sample that already exists, and it
converts a column from not-shippable to shippable. Nothing else on the tag
backlog has that ratio.

Found while running 2.2 (#1721).

ben-domingue:
## The gold labels are wrong, not the vocabulary

Before writing the `Mixed` sentence this issue asks for, I checked whether the tagger's answers were actually the errors. Mostly they aren't.

**1,239 of the gold-tagged tables carry a `cov_age` column**, so for a third of the tagged corpus the true age range is computable from the shipped data — no model, no source fetch, no new labelling. Audited a seeded random sample of 70 of them (`seed=1743`, `n=64` scoreable), comparing each human `age range` label against the ages in its own table:

| gold label | agrees with data | disagrees | |
|---|---|---|---|
[...]

Issue #1760, opened 2026-08-31
age range and sample: the tag vocabularies are enforced but never defined — two convention questions


ben-domingue:
@SamuelEnrique @saviranadela — two questions about tagging conventions that we
can't settle from the data, and you two are closest to these datasets.

Both are the same underlying issue: the tag vocabularies are now *enforced*
(the pipeline rejects a value outside the approved list) but they were never
*defined* — the list says what you may write, not what each value means. Where
two values could both apply, there's no rule, so different raters can be
consistent with the list and still disagree with each other.

## What we noticed

Many IRW tables carry a `cov_age` column, so for those we can compare the
`age range` tag against the ages actually present in the table. On a random
sample of 70 tagged tables that have `cov_age`, most labels line up
(`Adult (18+)` agreed 28 of 30). `Mixed` mostly doesn't: 26 of 29 disagreed,
and in 25 of those the table contains **no respondents under 18 at all** —
minimum age exactly 18.

Examples, all tagged `Mixed`:

| table | ages in the table | respondents under 18 |
|---|---|---|
| `colombia_2023_politics_voting` | 18 – 114 (n=46,392) | 0 |
| `mexico_2023_quality_wellbeingservice` | 18 – 98 (n=78,896) | 0 |
| `spain_2025_democracy_parties` | 18 – 97 (n=4,010) | 0 |
| `spain_2024_politics_beliefs` | 18 – 94 (n=2,562) | 0 |
| `margaretto_2025_translation_study_2_lextale` | 18 – 68 (n=220) | 0 |
| `silvia_2024_funny` | 18 – 88 (n=1,842) | 0 |

Mostly national survey programmes (CIS, ENSU and similar), plus a few lab
studies.

# Question 1 — `age range`

## The question

We don't think this is simply an error — there's a reasonable reading in
which it's correct, and which reading is right is a convention question
rather than a data question:

1. **The tag describes the shipped IRW table.** Then `Mixed` is wrong for
   these: the table holds only 18+, and someone filtering for `Mixed`
   expecting both children and adults gets a purely adult sample.

2. **The tag describes the source study or its target population.** Then
   `Mixed` may be right — a survey whose stated frame is broader can still
   release microdata filtered to 18+, and the tag is describing the study
   the data came from.

So: **when you tagged `age range`, which of those were you describing?**
And is that consistent between you, or did you approach it differently?

Samuel — several of the above are the Spain tables you processed, so you may
[...]

PR #1776, opened 2026-09-01
Adopt the #1760 decision rules for age range and sample

ben-domingue:
Draft for review — **nothing here is in force**. One file, `tags/proposals/1760_age_range_and_sample.md`, written to be read on its own: you shouldn't need #1760, #1743 or the tags sheet open beside it.

The RAs haven't replied to #1760. The question they were asked was about *intent*; setting the convention was always yours. This is written so the call can be made now and a later reply becomes a correction to a written rule rather than an answer to an open question.

**What it proposes**

- **`age range`** describes the shipped IRW table (option 1 in #1760), derived from `cov_age` where that column is usable, blank rather than guessed where it isn't. The deciding argument is the third one in §A0: it's the only referent that's *checkable*, so the column becomes derived and self-updating instead of a hand label that decays. Answers #1743 in passing.
- **`sample`** — the eight values are two facets sharing one column: setting (`Educational`/`Clinical`/`Program-based`/`Internet-based`) and frame (`Representative`/`Targeted/specific`/`General/non-specific`). Setting atoms combine freely; `Representative` and `Targeted/specific` are independent and often *should* co-occur; `General/non-specific` is the residual and never co-occurs with either.

**Measured impact** (from `metadata/tags.csv` and `metadata/metadata.csv`, 2026-09-01)

| | |
|---|---|
| `Mixed` tables adjudicable from their own `cov_age` | 472 of 700 |
| Untagged tables that gain an `age range` with no human effort | **1,120** |
| `sample` rows left exactly as they are | 2,299 of 2,480 |
| `sample` rows repaired | 181 |

**How to review it**

§*What I need from you* is six numbered decisions, each with the document's current position as its default — approving with no comment takes all six as written. The two I'd push on:

- **`MINOR_SHARE = 0.02`** (decision 2) is the one free parameter and the weakest thing here — a threshold between two failure modes, not derived from anything. Set it to `0` for a purely literal rule; none of #1760's cases change either way.
- **`Representative` + `Targeted/specific` co-occurring** (decision 3) killed my first draft, which used a precedence list where one frame value wins. 49 rows carry both, and a nationally representative sample *of teachers* genuinely is both. This is the difference between a definition and a re-tagging campaign.

Also in the file so it stands alone: the evidence restated from #1760, all seven of its example tables run through both rules with their current tags, the alternatives I rejected (including waiting), a per-rule confidence statement, a draft note to the RAs for your edit, and the reversal path.

https://claude.ai/code/session_01BUTbCySD4RLJrG7HGBMSz9


ben-domingue:
**All seven decisions are in, and the code is written, tested and dry-run. Nothing is published — the merge is the switch.**

## What is in the branch now

| | |
|---|---|
| `vocab.md` | Both rules inline — the referent, your `Mixed` definition, the two non-boundaries, the derivation and its guards, the two-facet split of `sample`. This is what the tagger reads. |
| `tags/decisions/1760_age_range_and_sample.md` | The reasoning, all seven decisions with who decided what, and the dry-run results. Moved out of `proposals/`; `vocab.md` links to it. |
| `metadata/tag_normalize.R` | Decision 3, enforced at export. |
| `metadata/03_tags.R` | Decision 7 — `age_range_derived.csv` outranks the Sheet for two columns only. |
| `tags/derive_age_range.py` | The derivation. Redivis-side aggregates, no downloads. |
| `tags/report_age_range_diff.py` | The dry run. |
| `metadata/11_status.R` | Per-column coverage (see the caveat below). |
| `metadata/tests/test_tags_union.R` | 45 tests, all passing. |

## The dry run

2,184 live tables carry a literal `cov_age`. **1,794 derive**; 345 are unusable and 45 quarantined.

Against the published column: **775 new rows** (tables nothing had tagged), **628 confirmed**, **382 changed**, 5 blanks filled, 4 `Non-human` preserved.

| from | to | tables |
|---|---|---|
| `Mixed` | `Adult (18+)` | 334 |
| `Child (<18y)` | `Mixed` | 32 |
| `Adult (18+)` | `Mixed` | 12 |
| `Mixed` | `Elderly (minimum age >50)` | 3 |
| `Adult (18+)` | `Child (<18y)` | 1 |

All seven tables named in #1760 resolve to `Adult (18+)`. `sample`: 181 rows repaired, `General/non-specific` 831 → 650, no other result set moves.

End-to-end simulation against the published `tags.csv`: 2,480 rows → 3,255, `Mixed` 700 → 526, `Adult (18+)` 1,260 → 2,125.

## Three things the dry run caught

1. **Below the 2% floor the tag fell through to `Adult (18+)` regardless of the majority** — `benitezsillero_2021_bullying`, ages 12–20 and 98% under 18, came out adult. It now takes the majority side, which is what the decision said.
2. **A derived age overrode `Non-human` on four tables.** Ages cannot contradict `Non-human`; `03_tags.R` now refuses it.
3. **A band shape the guard missed** — `alsyouf_2024_*` runs 1–7 with 7 distinct values, an age-bracket code that would have derived `Child (<18y)` for an adult sample. Also narrowed the months quarantine: where the maximum is ≤18 the tag is `Child` either way, so 91 of the original 104 held tables needed no holding.

## Two things for you, neither blocking

**Coverage will jump, and the headline would be wrong.** `status.json` counts a table as tagged if it has *any* row, so 775 age-only rows move it 55.3% → ~74% without anyone having tagged a construct or a sample. I added per-column coverage rather than let that number ship — today: age range 54.3%, construct type 54.5%, sample 53.9%. After publishing, age range goes to ~73% and the rest stay put. Worth deciding which number #1702 and the site quote.

**#1779** — the derivation found 81 live tables whose `cov_age` is a sentinel (999, 1999), a birth year, or a date offset (`-18090`). Not a tagging problem: shipped data that is wrong, and `irw_filter()` treats it as ages. Filed separately; it belongs to item 1's validator.

## What happens when you merge

I run the pipeline, post the after-action report on #1760 and close it, and close #1743. Nothing before that.

Issue #1777, opened 2026-09-01
Itemtext: ship the administered language, with English in parallel columns


xingyi-zhang:
@ben-domingue the automated skill has been shipping English for studies administered in another language. Manual processing already did this correctly, so this is an alignment, not a new idea.

## The precedent

`burkert_2019_whoqol_bref` ships the administered German — item text (`Haben Sie genug Energie für das tägliche Leben?`), German anchors (`Überhaupt nicht` … `Völlig`), and an instrument named `deutschsprachige Version`. That is the right call and it was made by hand.

The skill has not been following it. `ALSECYPIAMH_WU_2022_SDQ` ships canonical English SDQ wording for a Chinese administration; `altahla_2024_whoqol` ships English for an Arabic one. Both disclose it in provenance, but the administered wording is simply absent.

## Two problems, not one

1. **Where English was shipped, the administered wording is lost.**
2. **Where the original was shipped, no English is offered.** `burkert`'s 26 German items have no translation, so a non-German reader cannot use them.

Every curated itemtext table currently has exactly 10 columns, so there is nowhere for both to live.

## Proposed

Ship both. Administered wording in `item_text` / `option_text` / `instructions` / `section_prompt`; English in `item_text_translated` / `option_text_translated` / `instructions_translated` / `section_prompt_translated`; a `language` column naming the administered language.

`item` and `resp` are join keys and are never translated. `translated_substitute` becomes the fallback for when the original genuinely cannot be recovered, rather than the default.

## What needs your decision

The five new columns are **not in the public schema** at itemresponsewarehouse.org/itemtext.html. `validate_items.R` checks required columns are present and tolerates extras, so the gates are unaffected — but `upload.py` hands the CSV to Redivis, which infers fields from the file, so uploading such a table adds columns to `irw_text`. That needs agreeing before any upload, with `itemtext.html` updated in the same pass.

Also worth deciding: whether the existing tables get backfilled, or only new ones follow the policy.

## Live case

`weber2026_name_knowledge` in batch_015 (#1774) is the first table extracted under this rule — German administered, with the authors' own English rendering supplied in the same Dataverse record. It is ready but not uploadable until the schema question above is settled.

This also reopens the three `galindo_2023_*` tables I had scoped out of #1774 for being English-only: if their administered Spanish, Basque or Greek wording is recoverable, they stop being a liberal-bar call.

xingyi-zhang:
[...]

Issue #1774, opened 2026-09-01
Itemtext batch_015: extract the 13 tables freed by the Dataverse API sweep


xingyi-zhang:
Next extraction round, working the tables that #1751's Dataverse sweep just unblocked. Follows #1692 (batch_014) as the first full multi-table round.

## Why these

The sweep in #1771 moved 16 tables `BLOCKED` → `AVAILABLE` (audit-wide `BLOCKED` 214 → 171). None has item text, none is in `extraction_batches/queue_state.csv`, so none is claimed. This is work that did not exist two days ago, and the access problem is already solved — 27/27 records answered the API, with the recipe and the User-Agent gotcha documented in `itemtext/dataverse_api_route.md`.

Better still, each audit row already names the source artifact, so the slow judgment-heavy part of Step 3 is largely done. **13 tables across 8 records:**

| record | tables | what the record holds |
|---|---|---|
| `DVN/3CKJV1` | `weber2026_name_knowledge` | explicit `Item_wording_English.xlsx` — Item / Item wording table |
| `DVN/OXIMKR` | `boudreau_2022_police_reform` | `lawenf_1_codebook.docx` pairs each variable with full question **and every option** |
| `DVN/SNLKUE` | `balmas2018_leader_personality`, `balmas2018_leader_attitudes` | `Codebooks.pdf` maps every code to its administered question |
| `DVN/K0SRS8` | `lee2024_relative_clause` | `Appendix A.docx` is the administered test itself, verbatim |
| `DVN/5W9HXM` | `Adherence_Zissette_2018_SDB` | 24-page final structured survey with every question |
| `DVN/FXL7JW` | `gilbert_meta_39` | baseline + follow-up student questionnaires as administered |
| `DVN/8LBLYS` | `kushnir2017_anrt` | wording embedded in the data file's own column headers |
| `DVN/1KDJTZ` | `CFC_Balaji_2019` | headers pin `CFC1`–`CFC12`, the 12-item CFC scale |
| `DVN/IREEJJ` | `karajko2025_ai_governance`, `_ai_trust`, `_ai_risk` | `ESM_1.xlsx` headers are the administered question text |

Deliberately ordered by expected yield: the explicit-mapping and codebook records first, header-derived and instrument-PDF records after. Landing the clean ones first confirms the round end to end before the harder ones.

## Two things scoped in with eyes open

[...]

PR #1793, opened 2026-09-02
batch_015: item text for 11 tables freed by the Dataverse API sweep (#1774)

xingyi-zhang:
Closes the extraction round in #1774. **11 tables, all verified, none uploaded.**

Worked the records that #1751's Dataverse sweep moved `BLOCKED` → `AVAILABLE`. Files are in `itemtext/itemtables/batch_015/`.

| table | items | rows | how the mapping is pinned |
|---|---|---|---|
| `boudreau_2022_police_reform` | 4 | 20 | **exact reproduction** — source item × resp counts match live cell for cell |
| `kushnir2017_anrt` | 12 | 60 | **response-frequency matching** — label counts reproduce live integers cell for cell |
| `Adherence_Zissette_2018_SDB` | 26 | 156 | inline survey numbering + Marlowe-Crowne content check |
| `lee2024_relative_clause` | 18 | 36 | recovered header row + linguistic type codes |
| `balmas2018_leader_personality` | 14 | 70 | numbered codebook, per-item endpoint labels |
| `balmas2018_leader_attitudes` | 3 | 15 | same codebook |
| `weber2026_name_knowledge` | 59 | 295 | explicit wording file *and* source data headers |
| `karajko2025_ai_risk` | 10 | 50 | header code match |
| `karajko2025_ai_benefit` | 10 | 50 | header code match |
| `karajko2025_ai_governance` | 6 | 30 | header code match |
| `karajko2025_ai_trust` | 6 | 30 | header code match |

## Gates

`validate_items.R` 11/11 PASS against live · `audit_batch.R` **11 PASS, no anomalies** · `verify_batch.R` **PASS=11** · `normalize_nulls.R` clean · `lint_verification.R` **0 ERROR**

Every table has a re-runnable `verify_<table>.R`, driven by a shared `rederive.py` that rebuilds the expected text from the cached Dataverse sources — so the checks re-run rather than restating an evidence string. `verification_merged.csv` is included so `lint_verification.R`'s documented batch-dir invocation works.

The one lint WARN (`weber2026`, VERIFIED with a hedge) is the axis-scoping case you ruled on in #1696: VERIFIED covers item↔text, which the codes pin completely; the hedge is about resp *direction*.

## Four tables carry AI translation — held pending the indicator decision

`karajko2025_*` ship the administered **Bosnian/Croatian** in `item_text`/`option_text`/`section_prompt`, with English in the `*_translated` columns. That English is **machine translation produced by this pipeline**, not the study's and not human back-translated — no English version of those blocks exists in the record (`ESM_4.pdf` is entirely Croatian; the spreadsheet's English glosses cover only the unrelated P4 block).

Disclosed in provenance and in each `public_note`. A ready-to-paste `itemtext_issues.qmd` entry is held in `itemtext/fixes/itemtext_issues_karajko_translation.md`, **not applied** — pending the Slack discussion about how to mark AI-translated scales, and because `*_translated`/`language` aren't in the public schema yet (#1777).

One finding for that discussion: `item_text_translated`, `option_text_translated` and `instructions_translated` are **already shipping** in the curated corpus — 4, 4 and 2 tables respectively in a 22-table sample of 556, with column counts running 9–12 and `correct_response` absent from 16 of 22. So #1777 is narrower than it looked: documenting columns already in use, not introducing new ones.

## Three things worth a look at review

- **`boudreau`: resp 5 is "Don't know"**, not the scale extreme. Levels 1–4 fall monotonically (certif 850/372/167/49) then 5 jumps back to ~11% on all four items. In a `public_note`.
- **`kushnir`: "Don't know" was dropped rather than coded** — 8–16% of responses per item are absent, not missing at random. Its `item` join key is also the source column header, so it contains the item wording (37–107 chars). That's a response-table naming problem, not fixable here without breaking the join.
- **`lee2024`: a 25-vs-18 count mismatch, resolved rather than forced.** The processing script's own note says "original labels lost in Excel multi-row header"; they're in row 2 of the source xlsx, which numbers all 25 columns and places 1–18 on test items 1–18, excluding the O/X block.

## Not extracted, recorded in `pending_index_notes.csv`

- `gilbert_meta_39` — the 10 items are figural Raven-style matrices; no text exists. The audit's AVAILABLE verdict is right about the record and wrong about this table.
- `CFC_Balaji_2019` — one file of bare `CFC1`–`CFC12`, no codebook, no paper, empty dictionary row. Shipping it would mean assigning canonical wording by assumed order with no way to check.

Nothing uploaded — `uploaded` is empty on all 11 provenance rows.


ben-domingue:
@xingyi-zhang sounds like i need to upload some tables! where are they?

xingyi-zhang:
> @xingyi-zhang sounds like i need to upload some tables! where are they?

@ben-domingue  https://github.com/ben-domingue/irw/tree/main/itemtext/itemtables/batch_015

ben-domingue:
@xingyi-zhang done

Issue #1817, opened 2026-09-02
Two sessions backfilled the same population in parallel and neither knew


ben-domingue:
On 2026-09-01 two sessions worked #1777's administered-language backfill at the same time. Neither knew about the other. The work happened to be complementary — zero table overlap, verified — but that was luck, and the collision still cost three corrupted published tables and three contradictory numbers in the record.

## What happened

| | session A | session B |
|---|---|---|
| audit | tier A = 104 (#1795) | tier A = 83 (`69f28f9`, "the classifier was over-claiming") |
| built | 78 tier A tables | 3 tier B/C tables, wording recovered from the papers |
| plus | — | 17 tier B tables staged for the disclosure pass |
| uploaded | yes (#1813) | yes |
| landed in git | **yes** — PRs #1806, #1808, #1813 | **no** |

Session B worked on `itemtext/1777-backfill-audit`. The branch was never pushed and was then deleted. Its three commits — `8f35d6e`, `8f2cba2`, `6bd1171` — survived only as unreferenced objects reachable through one machine's reflog. A `git gc` would have taken them, along with the only record of three tables that are live on Redivis.

They were recovered on 2026-09-02 and pushed. Nothing was lost. But nothing made that recovery likely either — it happened because someone went looking for a directory that a stale issue said existed.

## What it actually cost

1. **Three published tables are doubled on Redivis** (#1816). #1813 fixed the uploader and repaired the 78 tables it could see. It could not see these three, so they are still corrupt today, five weeks of downloads later if nobody had looked.
2. **`brederecke_2020_sis` was shipped by one session after the other had ruled it un-shippable.** Session A's audit marks it `CANNOT BACKFILL`; session B split the composite field and uploaded it. Session B was right and the reasoning is good, but nothing arbitrated — the two records simply disagreed, and the live table settled it by existing.
3. **Three different tier A counts** are published across #1799, the README and the audit CSV: 104, 83, 79. Only one is a measurement.
4. **#1807 was filed one minute after PR #1806 merged**, describing a schema conflict as an open decision that the other session had already taken and executed. Anyone picking that issue up would have re-litigated a settled question and possibly re-done 78 tables of work.
5. **`verify_backfill.py` encoded one session's contract and failed the other's tables** for doing what they were supposed to do — base fields byte-identical vs. base fields legitimately rewritten to recovered wording. Two correct contracts, one gate, no note anywhere that there were two.

## What would have caught it

[...]

PR #1820, opened 2026-09-02
Record who produced a translation, via provenance translation_source (#1777)

xingyi-zhang:
Adds a provenance field recording where a `_translated` column's English came from.

The merged schema (#1794) treats `language` populated + `_translated` populated as one case, so a translation somebody published and one this pipeline generated are indistinguishable from the data.

New `provenance.csv` column `translation_source`, next to `text_source` because it complements it — a table is `text_source = study_materials` when the administered wording is the study's own, which says nothing about its English.

- `published` — this study's own supplement or deposit, or a validated English version of the instrument it cites
- `ai` — no published English existed, so this pipeline generated the translation
- empty — no `_translated` columns; administered in English

Two values, not three: a published translation is human by construction and this pipeline never hand-translates.

Applied to `batch_015` — the four `karajko2025_*` tables are `ai`, `weber2026_name_knowledge` is `published` (its `Item_wording_English.xlsx` is the authors' own), the other six empty.

`provenance.csv` is a repo record and is not uploaded, so this changes nothing on `itemtext.html` or in `irw_text` and needs no re-upload.

One gap it leaves open: the schema's issues-page requirement is scoped to its third row (`language` populated, `_translated` **empty**). The karajko tables are the second row, so nothing currently obliges a public callout for them. The held draft is in `itemtext/fixes/itemtext_issues_karajko_translation.md`.


ben-domingue:
@xingyi-zhang — this collides with something that merged a few hours before it, and the collision is worth knowing about before this lands.

**`translation_source` already exists on `main`.** PR #1815 added it to `itemtext/language_backfill/backfill_provenance.csv` with a four-value vocabulary, and it is populated on 78 rows:

| there (merged, 78 rows) | here (5 rows) |
|---|---|
| `official_instrument_english` — 14 live | `published` |
| `study_supplied` | `published` |
| `machine_translation` — 64 live | `ai` |
| `mixed` | — |

Same column name, same concept, two files, two vocabularies, neither session aware of the other. Both were written on 2026-09-02.

@ben-domingue's call: **keep the four values, remap these five rows.** Two reasons — it is 5 rows to change rather than 78, and the two-value set collapses a distinction the other file argues is real. Official instrument English is a *parallel version* rather than a translation of the administered string, so where a local adaptation diverges from the original, `_translated` stops corresponding word-for-word to the base field. Someone comparing the two columns should be able to tell which case they are in. Your reasoning for two values is sound in isolation; it is the 78 existing rows that decide it.

Concretely: `weber2026_name_knowledge` becomes `study_supplied` (its `Item_wording_English.xlsx` is the authors' own — that is exactly what the value is for), and the four `karajko2025_*` become `machine_translation`.

**PR #1822** makes the vocabulary single-sourced so this cannot happen a third time: the values move to `itemtext/provenance_vocab.csv`, `SKILL.md` and `language_backfill/README.md` point there instead of restating them, and `Rscript itemtext/check_provenance.R` exits non-zero on an unknown value — the way `TAG_VOCAB` already works for tags. Worth merging that first; then this is a five-cell edit.

## Your open question is already answered

You noted that `itemtext.qmd`'s issues-page requirement is scoped to its third decision-table row (`language` populated, `_translated` **empty**), so nothing obliges a callout for the karajko tables even though their English is machine-generated.

**That was settled this morning**, in a different thread, which is why it looked open: @ben-domingue ratified that machine translation ships with per-table provenance **and an issues-page line**, regardless of which row it sits in. So `itemtext/fixes/itemtext_issues_karajko_translation.md` should be applied.

The karajko tables turn out to be four of sixty. Your question exposed something bigger: `check_issues_page.R` globbed `itemtables/batch_*/provenance.csv`, so the backfill's 78 tables were invisible to it. Fixed in #1822, and the first run says **64 tables ship project-generated English and 60 have no public disclosure**. Those 60 entries are a separate mechanical pass, not yours to absorb here.

## One more, not a defect

`batch_015`'s `uploaded` being blank while `irw_itemtext()` reports the tables unavailable is not a missing step — the Redivis draft has not been released. That is the trap #1819 named today: red_up's `count(*)` verifies the *draft*, and everything users touch reads the *released* version. It resolves on release, and then the stamp is right.

ben-domingue:
@xingyi-zhang — `irw_text` was released as **v14.0** a few minutes ago, so the last item in your comment resolves: all 11 batch_015 tables are live and `irw_itemtext()` will report them available. `irw_text` now holds 579 tables.

**So `uploaded` can be stamped, and it is worth doing in this PR rather than a separate one** — this branch is the only thing currently touching `itemtext/itemtables/batch_015/provenance.csv`, and two branches editing the same provenance file is the collision that #1817 is about. One coherent edit here:

- `uploaded = 2026-09-02` on all 11 rows
- `translation_source` remapped to the four-value vocabulary: `weber2026_name_knowledge` → `study_supplied`, the four `karajko2025_*` → `machine_translation`

If you would rather not absorb the stamp, say so and I will take it after this merges — but not in parallel with it.

One thing the release does *not* change: the four `karajko2025_*` tables still owe an issues-page entry, because their English is project-generated. `itemtext/fixes/itemtext_issues_karajko_translation.md` is ready to apply. After #1822 lands, `Rscript itemtext/check_provenance.R` and `check_issues_page.R` will both report them as DUE until it is.

ben-domingue:
The conflict is `SKILL.md` only, and it is a clean one — #1822 landed a `translation_source` section at the same anchor. Resolution, in two steps:

**1. `SKILL.md` — drop your hunk entirely, keep `main`'s.** Yours describes the values inline; `main`'s now points at `itemtext/provenance_vocab.csv` and leaves the list in one place. Nothing of yours is lost: the *when to reach for each* guidance you wrote is what `provenance_vocab.csv`'s `meaning` column carries.

```
git fetch origin && git rebase origin/main
# on the SKILL.md conflict: take main's side (the block that says the values
# live in provenance_vocab.csv), drop the published/ai block
git checkout --theirs itemtext/.claude/skills/irw-auto-itemtext/SKILL.md   # during rebase, "theirs" is main
```

**2. `batch_015/provenance.csv` — remap and stamp.** No conflict here; it just needs the settled values, plus the upload date now that v14.0 is out:

| table | was | now |
|---|---|---|
| `weber2026_name_knowledge` | `published` | `study_supplied` |
| `karajko2025_ai_benefit` | `ai` | `machine_translation` |
| `karajko2025_ai_governance` | `ai` | `machine_translation` |
| `karajko2025_ai_risk` | `ai` | `machine_translation` |
| `karajko2025_ai_trust` | `ai` | `machine_translation` |

and `uploaded = 2026-09-02` on all 11 rows — the tables went live in `irw_text` v14.0.

`weber2026_name_knowledge` is `study_supplied` rather than `official_instrument_english` because its `Item_wording_English.xlsx` is *this study's own* rendering, not an instrument publisher's parallel version. That is exactly the distinction the four-value set exists to keep.

Then `Rscript itemtext/check_provenance.R` should pass on the vocabulary and report the four `karajko2025_*` as still owing an issues-page entry, which is the follow-up rather than this PR's job.

ben-domingue:
@xingyi-zhang — @ben-domingue asked me to unblock this rather than leave it waiting, so **I have rebased and updated your branch**. Force-pushed, so `git fetch && git reset --hard origin/xingyi/1777-translation-source` before you touch it again. Apologies for reaching into your branch; shout if you would rather I had not.

Three changes, all mechanical:

**1. Rebased onto main, resolving `SKILL.md` toward main.** #1822 landed after you opened this and moved the values into `itemtext/provenance_vocab.csv`, with `check_provenance.R` enforcing them. Your inline list was dropped in favour of that pointer — but nothing you wrote is lost: the *when to reach for each value* guidance is exactly what that file's `meaning` column now carries, and it is enforced rather than described.

**2. Remapped to the four-value vocabulary** (@ben-domingue's call, #1822):

| table | was | now |
|---|---|---|
| `weber2026_name_knowledge` | `published` | `study_supplied` |
| `karajko2025_ai_benefit` | `ai` | `machine_translation` |
| `karajko2025_ai_governance` | `ai` | `machine_translation` |
| `karajko2025_ai_risk` | `ai` | `machine_translation` |
| `karajko2025_ai_trust` | `ai` | `machine_translation` |

`weber2026_name_knowledge` is `study_supplied` rather than `official_instrument_english` because its `Item_wording_English.xlsx` is *the study's own* rendering, not an instrument publisher's parallel version. That distinction is the whole reason four values survived instead of two: a publisher's English is a parallel version rather than a translation of the administered string, so where a local adaptation diverges from the original, `_translated` stops corresponding word for word. Your two-value reasoning was sound in isolation; it was the 78 rows already using the four-value set that decided it.

**3. Stamped `uploaded = 2026-09-02` on all 11 rows.** `irw_text` v14.0 released them earlier today, which resolves the last point in your comment — `irw_itemtext()` reports them available now. I checked all 11 are live before stamping, so the date is a fact rather than an assumption.

## Your open question was already answered elsewhere

You noted that `itemtext.qmd`'s issues-page requirement is scoped to its third decision-table row, so nothing obliged a callout for the karajko tables. @ben-domingue ratified earlier that day that machine translation carries provenance **and** an issues-page line regardless of row — you could not have known, it happened in another thread.

That turned out to matter well beyond these four: `check_issues_page.R` globbed only `itemtables/batch_*/provenance.csv`, so the language backfill's 78 tables were invisible to it, and **60 tables were shipping IRW-generated English with no public disclosure**. Fixed in #1822, disclosed in datapages/irw#112. Your four are in **datapages/irw#114** — and the held draft at `itemtext/fixes/itemtext_issues_karajko_translation.md` turned out to be written in the `.callout-warning` format the page retired on 2026-09-01, so it is carried into the YAML shape rather than pasted.

With that merged, `check_provenance.R` reports 0 machine-translated tables without an entry.

Issue #1709, opened 2026-08-29
[7] Scale item text - the asset nothing else has


ben-domingue:
Item **7** of the Year 3 roadmap. Tracking issue: #1702 &nbsp;·&nbsp; [full context](https://claude.ai/code/artifact/a322f671-38af-48c1-b4c8-1e80cbf8ec51#i7)

167 of 1,300 tables (12.7%) have item text; 805 (61.5%) are known from human annotation to have text available. A 4.8x expansion of an asset no comparable resource holds, with tooling already built and 8 open ITEMS issues mapping the work. Reach argument: the IRT-for-LLM-benchmarks literature (PSN-IRT, IRT for AI safety, fluid benchmarking) has thousands of items against fewer than 100 examinees — item text plus human responses at IRW scale is exactly what it lacks.

---

### Corrected 2026-08-31

Recounted from `itemtext/availability_audit_full.csv`, `itemtext/extraction_batches/queue_state.csv` and `metadata/itemtext_metadata.csv` (regenerated 2026-08-31). The figures above are left as written.

| | as written | measured 2026-08-31 |
|---|---|---|
| tables with item text | 167 of 1,300 (12.7%) | **561** (558 of them live) |
| audit population | 1,300 | **2,110** rows, 2,107 still live |
| classified `AVAILABLE` | 805 (61.5%) | **1,349** (64.0%) |
| classified `BLOCKED` | 215 | **214** |
| extraction queue pending | 1,164 | **1,165** (122 done, 54 excluded, 11 failed) |

**The 12.7% needs a denominator before it is requoted.** As written it is `167 / 1,300` — a share of the *availability-audit* population, not of the corpus. #1702's headline table carries the same 12.7% in a row whose neighbours are all denominated in live tables, so the two documents currently disagree about what is being measured. Today the two framings are **558 of 4,134 live tables (13.5%)** and **558 of 2,110 audited (26.4%)**. The reach argument is unaffected either way — the target grew from 805 to 1,349 — but the multiplier changes with the choice, so state it.

`BLOCKED` at 214 vs. the 215 above vs. 218 in the linked write-up is drift across three snapshots of the same file, not three different findings. **2.5c** (#1765) is where these should come from.

### Sub-actions
[...]

Issue #1812, opened 2026-09-02
Item text column order: the schema says what the columns are, not what order they come in


ben-domingue:
Two batches now write the administered-language columns in different orders:

- **`batch_015`** (#1793): `... resp, item_text_translated, option_text_translated, instructions_translated, section_prompt_translated, language` — `language` last.
- **the tier A backfill** (#1799): `... instrument, language, instructions, ... resp, instructions_translated, section_prompt_translated, item_text_translated, option_text_translated` — `language` after `instrument`, matching the field order in `itemtext_standard.md`.

Neither is wrong, because nothing enforces an order and Redivis infers fields by name. Ben's call (2026-09-01): **not critical, and column order should not be presumed anywhere.**

Worth doing anyway, cheaply: `itemtext_standard.md` lists the fields in a table, which reads like an order without saying it is one. Either state explicitly that order is not significant, or state the canonical order and have `validate_items.R` check it. Right now the document is ambiguous enough that two batches read it two ways in the same week.

Recording it because it currently lives only in a merge-commit message on #1806.

Related: #1777 · #1793 · #1799

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0152893Dc4fTon7NoNdEYXSa

ben-domingue:
Folding this into the doc-consistency one-off alongside #1729, #1730 and #1731 — all four are one-line edits to documents that read as authoritative and are not. Ben's 2026-09-01 call stands (column order is not critical), so the work here is the second half of the body: state explicitly in `itemtext_standard.md` that field order is not significant, so the table stops reading like an order.

Triage pass 2026-09-06.

PR #1848, opened 2026-09-03
itemtext: ENEM 2023, four tables at full coverage (#1709)

mateusmazza:
Item text for `enem_2023_1mil_{lc,ch,cn,mt}`: **184 items, 920 option rows, 100% of the items in the live response tables** (MT is 44 because #1942 dropped annulled item `14887`).

This is the pilot year. If the approach is accepted, the same workflow repeats for the other 12 ENEM years.

Refs **#1709** (scale item text). It does **not** close it: 4 of the 52 `enem*` tables are done.

Everything is in `itemtext/itemtables/batch_enem_2023/`: the four `__items.csv`, `notes.csv`, `provenance.csv`, `verification_merged.csv`, `audit_report.csv`, and `scripts/` (the DOSVOX parse, the PDF-sourced items, and the position→item join).

## Source

INEP's own **accessibility booklet** (LARANJA — Braille / Adaptada Ledor), read from the DOSVOX screen-reader plain text shipped inside the 2023 microdata. INEP wrote verbal descriptions of figures, graphs and tables into that booklet, so figure content is INEP's wording. That matters for an exam that relies this heavily on figures.

## Join

Booklet position → IRW `item` via `ITENS_PROVA_2023.csv` (`CO_POSICAO` → `CO_ITEM`), using the regular-application LARANJA booklet per area (CH 1197, LC 1207, MT 1217, CN 1227). `item` values are the publisher's own. `mapping_basis=paper_explicit`, `route=publisher_position_item_map`.

LC's 50 items over 45 positions are the foreign-language block, split by `TP_LINGUA`.

## Generated text

Generated text is confined to `item_text` / `item_text_translated` and always marked inline `(gerada por IA)` / `(AI-generated)`. **No `option_text` cell is generated.** CN and MT carry `description_source=partly_generated` (owes an issues-page entry at upload); CH and LC are `source_published`. The rule and its check are #2165.

| Item | Table | What is generated |
|---|---|---|
| `78578` | cn | absent from the accessibility booklet, transcribed from the AZUL PDF; equation, table and option-format descriptions in `item_text`. Printed options are bare graphs, so **`option_text` is blank** |
| `54804` | cn | absent from the booklet, from the AZUL PDF; structural-formula descriptions in `item_text` |
| `60332` | cn | in the booklet, but INEP declined to describe the figure; that one sentence is replaced |
| `125902` | mt | absent from the booklet, from the AZUL PDF; graph and option-format descriptions in `item_text`. Printed options are bare diagrams, so **`option_text` is blank** |
| `81742` | mt | absent from the booklet, from the AZUL PDF; pentagonal-number figure description in `item_text` |

Every key was checked against the PDF independently of the position map. `correct_response` and `resp_raw` keep `78578` and `125902` addressable even without option wording.

## Language (#1777)

`language=Portuguese`, with English machine translation in the four `_translated` columns (`translation_source=machine_translation`). Issues-page disclosures (four for machine translation, plus CN and MT for generated descriptions) are owed in the same round as the upload (separate PR against `datapages/irw`).

Stimuli already in English or Spanish are the object of study and are kept verbatim rather than round-tripped, as are words the options analyse metalinguistically.

## `resp_raw`

`resp_raw` (spelling per #2094) carries the printed option letter A–E alongside the scored `resp` (0/1), so `option_text` joins to the letter a respondent chose. Without it the four distractors are anonymous.

## `section_id` / `section_prompt`

Trivial `<table>_1` and empty. No ENEM item shares a passage with another, verified four ways.

## Gates (on the current head)

| Gate | Result |
|---|---|
| `validate_items.R --table-sets` (live, server-side) | **PASS ×4**, item and resp sets |
| Answer key vs corrected response CSVs (`resp_raw` scored 1 = `correct_response`) | 0 disagreements across 184 items |
| `audit_batch.R` (live) | 1 PASS, 3 WARN, all explained in `notes.csv` |
| `irw-validate` | ok ×4 |
| `normalize_nulls.R` | idempotent |
| `lint_verification.R` | no problems |
| `check_provenance.R` vs site `main` | exit 0 (tables not yet uploaded, so disclosures not yet enforced) |

The three audit warnings are properties of the exam, not defects:

- **CN**: item `66330`'s printed options *are* the numerals 1–5, so `option_text="1"` collides with `resp=1` by coincidence. Also, `78578` has no `option_text` (bare graphs, above).
- **MT**: `125902` has no `option_text` (bare diagrams, above).
- **LC**: the row-count anomaly is exactly the ten foreign-language items: **577,679 English + 422,321 Spanish = 1,000,000**.

https://claude.ai/code/session_01Hyv9hLNt2ShNdhifJ7DjZ2


ben-domingue:
@mateusmazza really nice. this dovetails with some separate convos that @xingyi-zhang and i have been having about translation. i wanted to try to make a few 'rules' here governing translations but i'm open to whether they are sensible. an exec summary here and then more below:

Really nice work, and the sourcing decision is the best part of this — using the LARANJA accessibility booklet so the figure descriptions are INEP's own words rather than ours is exactly right for an exam this figure-heavy, and I don't think I'd have thought of it. The `ITENS_PROVA` position→item join means `item` is the publisher's, which clears the usual worry about bare-integer item codes. 180 of the 185 items look clean to me.

Three things I'd like to talk through before we merge, mostly because this is the pilot for 12 more years and I'd rather settle the rules once.

**1. Generated option text.** I'm comfortable with the AI-written figure descriptions inside `item_text` for the five items INEP didn't describe — that's an annotation on a stimulus, clearly marked. I'm less comfortable with the generated `option_text` on `78578` and `125902`, where the printed options are bare graphs. A generated option label is something people will join to `resp` and analyse as if it were printed on the page. My inclination is to blank `option_text` on those two items and keep `correct_response` and the option letters, which still leaves them addressable. Does that lose anything you were counting on?

Related and smaller: the inline marking isn't quite complete — the 25 `item_text` cells carry `(gerada por IA)`, but none of the `option_text` cells do, including the ten that are entirely generated. Could you give me a per-cell list of what's generated (item + column) rather than the prose summary? That's the part I can't check myself.

**2. `raw_resp`.** I think you've convinced me on the substance — with `resp` at 0/1 the four distractors really are anonymous otherwise, and the nominal tables do carry the letter. But no itemtext table has used this column before, and we're now spelling the same idea three ways: `resp_raw` in the response tables and on the public schema page, `raw_resp` here, `text` in the nominal tables. Any objection to picking one and my documenting it properly? I also need to fix `validate_items.R`, which currently skips the resp-set check whenever `raw_resp` is present — that quietly weakened the gate on your tables.

**3. Reproducibility.** There's no build script in the PR. For one year that's fine; for thirteen I'd really like the DOSVOX parse and the position→item join committed, the way you did the FarmShare run kits. Is that easy to pull out?

Two small mechanical things I can fix myself if you'd rather: the four new `pending_index_notes.csv` rows are missing the `status` field, and `check_provenance.R` reported clean only because my local site checkout was sitting on a branch — the four issues-page disclosures are still owed. And one bit of housekeeping on my end: our own docs say ENEM item text is "handled by Ben separately," which is why the automated pipeline skips all 52 tables. That should say you.

ben-domingue:
The longer version, in four parts: the detail behind the three points above, the translation rules I'd like to propose, the remaining smaller findings, and what I'd suggest as standing terms for years 2–13 so we don't relitigate any of this twelve more times.

---

## 1. Generated text — where the line goes

The distinction I'm drawing is between **annotation** and **invention**.

A generated description of a figure sitting inside `item_text` is an annotation on a stimulus that exists. The reader can see it is a description, it is marked, and the underlying stimulus is still the exam's. That is a reading aid, and for an exam where a large fraction of items are unreadable without one, it is worth having.

A generated `option_text` is different in kind. It becomes a *label on a response category*. Someone doing distractor analysis joins it to `resp`/`raw_resp` and treats the string as what was printed on the page — and there is no natural moment where they stop and ask whether the label was authored by us. The `(gerada por IA)` marker helps a human reader and does nothing for a `merge()`.

So: `78578` and `125902` lose their `option_text` and `option_text_translated`, ten rows. `correct_response` (B and A, from INEP's own `ITENS_PROVA`) stays, and `raw_resp` stays, so the options remain addressable and joinable — you just can't read them. `pending_index_notes.csv` and `public_note` should say the printed options are bare graphs/diagrams with no text at all, which is a fact about the exam and worth recording.

On the marking gap: 25 rows carry `(gerada por IA)` in `item_text` and `(AI-generated)` in `item_text_translated`, on the five items `54804`, `60332`, `78578`, `81742`, `125902`. Zero rows carry either marker in `option_text` — including the ten that are wholly generated. I'm not reading that as an attempt to hide anything; the PR text is explicit that those options are generated. But it means "all generated text is marked inline" isn't a property I can check by grepping, and the prose summary and the file disagree. What I'd like is a per-cell inventory — item, column, and what the source was — for every cell that isn't a transcription. If the answer is "those ten rows and nothing else," great, that's a one-line answer and item 1 disposes of it.

**Proposed rule, for the repo not just for you:** generated descriptions are permitted in `item_text` only, always inline-marked, never in `option_text`, and a table carrying any gets `text_source=partly_generated` in its provenance plus an entry on the public issues page — the same obligation `machine_translation` already carries. I'd add `partly_generated` to `provenance_vocab.csv` and have `check_provenance.R` enforce both halves, including a hard check that `option_text` never contains a marker. We learned on #1820 that a vocabulary living in prose gets two incompatible implementations inside a week, so this should be a thing that exits non-zero.

## 2. Translation rules

This is the part I most want reactions to, from you and from @xingyi-zhang, because your PR quietly invents a rule I think is right and that we haven't written down.

The contract as it stands from #1777:

- **R1.** Base text fields hold the wording respondents actually read, verbatim. English goes in the four `_translated` columns. `language` names the administered language.
- **R2.** If the study, its deposit, or its own supplements published an English version, ship that. Only if none exists do we translate — and then `translation_source=machine_translation`.
- **R3.** Every `machine_translation` table gets a line on the public issues page, saying in plain words that the English is ours and unreviewed. Ratified 2026-09-02.

What this PR adds, and what I'd like to adopt:

- **R4 (new, yours).** *A stimulus that is itself in another language, and is the object of study, is kept verbatim in the `_translated` column rather than round-tripped.* The LC foreign-language items are the clean case: translating an English Donne poem from its Portuguese-exam context back into English produces a different poem, and the item stops being answerable. Same for the options that analyse a specific Portuguese or Spanish word metalinguistically — the word stays, with a gloss. This generalises well past ENEM: any language test, any item where the wording *is* the construct.
- **R5 (new, yours).** *`language` describes the administration, not the cell.* All 250 LC rows read `Portuguese` even where the stimulus is English or Spanish, because a Brazilian student sat a Portuguese-language exam. I think that is right and it is consistent with how we already use `language` in the translated-substitute fallback, where `language=Chinese` sits next to English `item_text`. Worth stating outright, because the naive reading is per-cell and someone will eventually "fix" it.

One thing R4 does need: a way to tell "kept verbatim on purpose" from "forgot to translate." Right now they are byte-identical. The cheapest fix is that the provenance note names the items, which yours does. If we expect this often, a marker or a per-item flag would be better — I don't have a strong view and would rather hear yours.

## 3. Smaller things

- **`mapping_basis=data_labels` is the wrong value.** Your mapping is a publisher-supplied position→item table, which is `paper_explicit`. This isn't pedantry: `data_labels` is precisely the value that exempts a table from the Step 5b independent mapping check and from `lint_verification.R`'s `NOT_NEEDED` rule. Your evidence in `mapping_verification.csv` is strong enough to carry `paper_explicit` on its merits — the four-way booklet cross-check and the five re-derived keys are more than most tables have — so it doesn't need the exemption, and taking it makes the record say something untrue about how the mapping was established.
- **Batch layout.** The four `__items.csv` are at `itemtables/` top level while their `provenance.csv` is in `itemtables/batch_enem_2023/`. `BATCH_PROCESS.md` expects `{table}__items.csv`, `notes.csv`, `provenance.csv`, `verification_merged.csv` and `audit_report.csv` all in the batch dir; as committed, `lint_verification.R` errors on a batch dir with no `verification_merged.csv`, and `audit_batch.R` would write its report into a directory with nothing to audit.
- **`route=publisher_position_item_map`** is a new route name outside the numbered-route vocabulary. I think it deserves to exist — it is a genuinely distinct and rather strong route — but let's add it deliberately rather than have it appear.
- **`check_provenance.R`.** Worth knowing why its clean exit didn't mean anything: the script reads the issues page from the sibling `irw_site` checkout, and refuses to enforce the disclosure check when that checkout isn't on `main`. Mine was on a feature branch, so it reported and did not enforce. The page currently has no `enem` entries at all. Four disclosures owed for the machine translation, and under the proposal in part 1, two more for the generated descriptions.
- **`pending_index_notes.csv`** — the four new rows have two fields against a three-field header; the `status` value is missing.

## 4. Things I checked and am explicitly *not* asking you to change

Listing these so they don't come back as review comments in year two:

- **Column order.** Yours matches `batch_015`; the tier-A backfill uses a different order. I ruled on #1812 that order isn't significant and shouldn't be presumed anywhere. Non-issue.
- **Trivial `section_id` / empty `section_prompt`.** Your four-way check — no shared-passage markers, no question lacking its own stem, no adjacent stems above 45% similarity, no long sentence in two stems — is more than I'd have asked for. Correct as shipped.
- **Gating with `validate_items.R --resp-csv` against local 45M-row CSVs** rather than live Redivis. That is the quota-safe route and the right call; `irw_fetch` on these tables would be a bad day for the export budget.
- **Annulled item `14887`** — blank `correct_response`, all five options at `resp=0`. That mirrors what `data/enem_2023.R` actually computes, since no response letter can equal `X`. Exactly right, and it's the template for the other nine annulled items across the years.
- **The `[supressão de texto]` items.** INEP's own suppression of copyrighted stimulus, transcribed as INEP marked it. That is the source being honest and us not papering over it.

## 5. Standing terms for 2013–2025

If these look right to you, I'll write them into `BATCH_PROCESS.md` so the next round doesn't rediscover them:

1. The accessibility booklet is the primary source every year. The AZUL PDF is a fallback only for items the booklet substitutes away, and the provenance says which.
2. Generation is confined to `item_text`, always inline-marked, and the table carries `text_source=partly_generated`.
3. One batch dir per year, `itemtables/batch_enem_<YYYY>/`, with the full five-file set.
4. The run kit lands once, not once per year.
5. Each year's four disclosures go up in the same round the tables do.
6. Annulled items follow the `14887` handling.

And housekeeping on my side regardless of any of the above: `SKILL.md` and `BATCH_PROCESS.md` say ENEM item text is handled by me, which is why all 52 rows sit `excluded` in `queue_state.csv`. It's you, it arrives by PR, and those four rows need a terminal status that means "shipped outside the pipeline" rather than one that means "never do this."

santiagorq2405:
Two things from the side of someone who would join `option_text` to the nominal tables, following Ben's review above.

On the naming question. `raw_resp` already exists in the response tables too: seven core tables carry it, and it is the one spelling of the three that no standard page documents.

- `raw_resp` (7, undocumented): `fullscaleiq_memory`, `fullscaleiq_mentalrotation`, `fullscaleiq_vocab`, `much_tte_2025_concentrationtask`, `much_tte_2025_matrixreasoning`, `wilmer-mrmet-normative-data-set-2022`, `wilmer-rmet-normative-data-set-2022`
- `resp_raw` (6, on `itemtext.html`): `borges_brazil_residency_2024_cbt`, `environment_ltm`, `himmelstein-berlin_numeracy-2025`, `mentalrotation_wolf_2024`, `nomt_hooper_2024_study2`, `science_ltm`
- `text` (1, on `nominal_standard.html`): `borges_brazil_residency_2024_pbt`

No table carries two of them. Counted on the live catalogue on 2026-09-04. So whichever name wins, the scope is fourteen tables, and the seven undocumented ones need a change either way, a rename or a doc entry.

On `78578` and `125902`, and whether blanking `option_text` loses anything: nothing on the consumer side. Anyone grouping distractors by their text would pull a generated label into the grouping without ever seeing the marker, which is the `merge()` point exactly.

mateusmazza:
@ben-domingue 

Rebased and brought up to date. Five things changed:

**1. Dropped item `14887` from MT.** #1942 removed items INEP annulled after the exam, so the live `enem_2023_1mil_mt` is 44 items, not 45. Its five option rows are gone. This would have failed the item-set gate on upload.

**2. `raw_resp` → `resp_raw`** in all four tables, per #2094. No gate depended on the old spelling.

**3. Moved into `itemtables/batch_enem_2023/`** with the full five-file set (`notes.csv`, `verification_merged.csv`, `audit_report.csv` are new). The four CSVs are pure renames. Kept the name `batch_enem_2023` rather than taking `batch_203`, since the queue runner allocates numbers sequentially and would collide.

**4. `mapping_basis`: `data_labels` → `paper_explicit`.** I had this wrong. Step 5b says the exemption comes from the code derivation, never from `mapping_basis`, and `data_labels` claims the *words* came from data labels — ours come from INEP's accessibility booklet. Nothing was relying on the carve-out, so no `verify_*.R` is owed.

**5. `queue_state.csv`:** the four rows go `excluded` → `done`. The other 48 `enem*` rows stay `excluded`.

**Re-validated against the corrected tables**, not the pre-fix CSVs:

```
lc  PASS item sets match exactly (50 items)  PASS resp sets
ch  PASS item sets match exactly (45 items)  PASS resp sets
cn  PASS item sets match exactly (45 items)  PASS resp sets
mt  PASS item sets match exactly (44 items)  PASS resp sets
```

`lint_verification` 0 ERROR, `check_provenance` exit 0, `normalize_nulls` idempotent. `audit_batch` gives 2 PASS and 2 WARN; both WARNs are structural and explained in `notes.csv` — CN item `66330`'s options genuinely are the numerals 1–5 (the stem asks which of five numbered tests causes the scaling), and LC's 10 flagged items are the elective blocks, 422,321 + 577,679 = exactly 1,000,000.

One correction to your review: `validate_items.R` does **not** skip the resp-set check when `raw_resp` is present — the branch keys on whether `resp` is populated, and ours is 0/1, so the check ran on all four. Same in `audit_batch.R`. Your underlying point about the three spellings was right, and #2094 settled it.

Also fixed the skill's standing exclusion, which said you were handling ENEM item text.

**Could you do a final check before I run the other 12 years?** 2023 is the pilot and the per-year work is mechanical from here, so if the shape of this batch is wrong I'd rather change it once than 12 times.

ben-domingue:
@mateusmazza thanks, this is in good shape. I checked it and pushed the last fixes to the branch myself, so there's nothing for you to redo on 2023.

**What I checked.** Live item and resp sets match on all four tables. Using the corrected response CSVs, the letter scored 1 matches `correct_response` on all 184 items. I also solved nine random items by hand, and every key is right. You were right about `validate_items.R`: the resp-set check did run.

**What I changed:**
1. Blanked `option_text` / `option_text_translated` on `78578` and `125902`, since those were generated descriptions of bare graphs and diagrams. `03_join_2023.py` now does this itself and emits `resp_raw`.
2. Restored the line endings in `pending_index_notes.csv` and `mapping_verification.csv`. They had been rewritten CRLF→LF, which turned a four-row addition into ~1,800 changed lines.
3. Fixed stale notes: `raw_resp` in all four, 14887 still described as shipped in MT, and the five-vs-four PDF item count.
4. Added `description_source` to `provenance.csv`: `partly_generated` for CN and MT, `source_published` for CH and LC.

**Rulings, now written into `BATCH_PROCESS.md` (#2165):**
- **Generated descriptions** go in `item_text` only, always marked, never in `option_text`. The table carries `description_source=partly_generated`, which owes an issues-page entry like machine translation does. `check_provenance.R` now enforces this against the item files.
- **R4** (a foreign-language stimulus that is the object of study stays verbatim in `_translated`) and **R5** (`language` names the administration, not each cell): adopted as you did them.
- **Standing terms for the other 12 years** are there too. Two changed from my 09-03 list:
  - The accessibility booklet is primary *where a year has one*. Your script notes some years are PDF-only.
  - Annulled items are simply not shipped, since #1942 removed them.
  - Also, whatever writes the two shared CSVs should append rows rather than rewrite the file.

**One question before we upload**, since it sets the pattern for all 13 years. INEP's microdata page carries the gov.br footer: *"Todo o conteúdo deste site está publicado sob a licença Creative Commons Atribuição-SemDerivações 3.0 Não Adaptada"* (CC BY-ND 3.0). We ship English translations, and a translation counts as a derivative work, so I want to know whether that footer actually covers the exam files.
1. Does the microdata zip, or the provas/gabaritos pages, carry its own licence or terms of use, separate from the site footer? A LEIA-ME, a PDF preamble, anything that names a licence?
2. Does INEP say anything about reproducing exam items specifically? For example, that provas are public government works, or any conditions on reuse?

Nothing to change in the tables. I just need INEP's own words, with a link or file name, so the decision rests on a quote.

ben-domingue:
@mateusmazza one more round on 2023 before you start the other years. An outside check against ITENS_PROVA and the DOSVOX files turned up three things, now fixed in #2169, and they change the standing terms in `BATCH_PROCESS.md`:

1. **The accessibility booklet swaps items rather than dropping them.** LARANJA CN carries 14385/14397 in place of 54804/78578, and MT carries 14435/141552 in place of 81742/125902. Your tables were right, but each year, compare the accessibility `CO_PROVA`'s item set with the regular one before joining by position. The difference is exactly the items that need the PDF.
2. **Check `IN_ITEM_ABAN`, not just `TX_GABARITO`.** MT 86360 is flagged abandoned ("Bis<0,01") but keeps key D. We're keeping items like that and naming them in the notes; only items with no valid key are dropped.
3. **The DOSVOX wording isn't the printed booklet's.** It spells out notation ("gramas por mol", "Q índice 2") and moves credits. Only 2 of 44 MT and 5 of 45 CN items match AZUL word for word. We're shipping it as is and saying so in every table's `note` and `public_note`, so please carry that disclosure each year.

Also, the "soletrar" screen-reader line is stripped from `instructions`. Nothing else for you to redo on 2023.

Issue #307, opened 2024-09-12
Network structure of the Wisconsin Schizotypy Scales–Short Forms: Examining psychometric network filtering approaches


ben-domingue:
https://link.springer.com/article/10.3758/s13428-018-1032-9
https://osf.io/c6rqy/

aanyatash:
Hi @ben-domingue, I worked on converting this dataset. Please let me know what you think – thanks!

[christensen_2018_wsssf.zip](https://github.com/user-attachments/files/31666044/christensen_2018_wsssf.zip)

ben-domingue:
one question: there seem to be some respondents responding to the same item multiple times (just looking at the numbers of `id` and `item` values and the number of rows). is there any explanation offered about that? @aanyatash

aanyatash:
Hi @ben-domingue, each respondent has all 60 items with no repeats as rows / (ids × items) = 1 for both files (349,860 = 5,831 × 60 and 130,260 = 2,171 × 60). Every participant's item count should be 60 (min = max = 60). Not sure which rows you saw as duplicated, so please let me know which to look at if I'm missing something – thank you!

ben-domingue:
@aanyatash my mistake, you are right. ship pr/dict/tags? thank you!

aanyatash:
Hi @ben-domingue, just added tags, dict, and pr (https://github.com/ben-domingue/irw/pull/1881)!

PR #1881, opened 2026-09-03
Network structure of the Wisconsin Schizotypy Scales–Short Forms: Exa…

aanyatash:
…mining psychometric network filtering approaches

 #307


aanyatash:
Hi @ben-domingue! To clarify, I should be calling run_qc or validate_frame in the script to pass this check? Thanks!

ben-domingue:
@aanyatash oh, hm. i think the agent is doing something i hadn't anticipated. so it is now triggering an error? let me look into this. sorry!

PR #1920, opened 2026-09-04
concretewords: put the rated expression on the item axis, not the rater

### LINKED ISSUE #1876: concretewords: item and id axes are transposed (item holds respondent IDs)
`concretewords` has its `item` and `id` axes the wrong way round.

`data/concretewords.R` assigns:

```r
item <- x$Participant
id   <- x$Expression
```

So the IRW `item` axis holds **anonymous Qualtrics respondent IDs**, and the rated multiword expressions — the things that are actually the items — sit in `id`.

## Evidence

From server-side aggregates (no export):

- 1,831 distinct `item` values
- **1,831 / 1,831** match the Qualtrics ResponseID pattern `^R_[A-Za-z0-9]{15,17}$`
- **0 / 1,831** contain a space, which is impossible for a corpus of English *multiword* expressions
- First values: `R_036k0LpyK0SQ68p`, `R_09E2jMmul7dLASl`, `R_0CfyKBsf3z8JKMh`

For contrast, if `item` were the rated expression you would expect ~62,000 items, nearly all containing a space.

## Consequence

Beyond the axes being semantically backwards, this makes item text **permanently unattachable**: an anonymous respondent ID has no stem, so no extraction pass can ever fill `item_text` for this table. If the table were re-derived with `item = Expression`, item text would be trivially available, because the expressions are self-describing.

## Reproducing

`itemtext/itemtables/batch_016/verify_concretewords.R`, quota-safe.

## Related

`itemtext/availability_audit_full.csv` classifies this table `AVAILABLE`, reasoning that "the openly published norms dataset is the actual list of rated expressions". That is true of the *study* and false of the IRW *item axis*; that row should be revised whichever way this issue is resolved.

ben-domingue:
Fixes #1876.

`data/concretewords.R` had the two axes the wrong way round:

```r
item <- x$Participant   # anonymous Qualtrics ResponseIDs
id   <- x$Expression    # the rated multiword expressions
```

The IRW `item` axis held 1,831 **raters**, and the ~62,000 rated expressions — the things that actually are the items — sat in `id`. Beyond being semantically backwards, this made item text **permanently unattachable**: a respondent ID has no stem, so no extraction pass could ever fill `item_text`.

Swapped. Each `item` value is now the expression itself, so the item code carries its own wording — the same self-describing pattern as `kushnir2017_anrt`.

## Verification

No `Ratings_RawData.csv` is cached in the repo, so this was verified by running the **actual script** — from `origin/main` and from this branch — against a synthetic source with the real columns (`Participant`, `Expression`, `Rating`, `Filter`), including two rows that must be dropped: one `Filter == 0` and one `Rating == 6` ("I don't know the meaning of this expression").

| | distinct `item` | distinct `id` |
|---|---|---|
| before | 4 — `R_oHwLMeaZqo9DZDNj` … | 5 — `ice cream sandwich` … |
| after | 5 — `ice cream sandwich` … | 4 — `R_oHwLMeaZqo9DZDNj` … |

- row count preserved
- the two axes are an **exact swap** (`old$item == new$id` and `old$id == new$item`)
- `resp` identical
- both filters still drop exactly the rows they should

## Also in this change

Both called for by the issue:

- **`itemtext/availability_audit_full.csv`** — the `concretewords` row read `AVAILABLE` on reasoning that was true of the *study* and false of the IRW *table*. The classification is now correct, but for the right reason, so the reasoning is rewritten. **Edited in place**: neither a `QUOTE_ALL` nor a `QUOTE_MINIMAL` round-trip of that file is byte-identical, so it must not be reserialised. Diff is one line; the file still parses at 2,111 rows.
- **`itemtext/fixes/concretewords__items.PARKED.README.md`** — records that the parked file is superseded and must **not** be promoted, since its item axis is the old one. Its `instructions` / `option_text` / `instrument` fields survive the fix and are reusable verbatim; only the item axis has to be regenerated.

`verify_concretewords.R` is left unchanged as the batch_016 record of what was found, with a banner noting it asserts the broken orientation on purpose and will correctly report `ok=FALSE` once the corrected table is live.

## Before merging / uploading

- **Not uploaded.** The live table still serves the swapped axes. This re-derives the item axis from 1,831 values to ~62,000, so it is a **new version via the draft-release path**, not a correction in place — the largest blast radius of the four `data fix` issues.
- Synthetic data proves the transformation, not that the raw file's columns are named as the script assumes. `Participant` / `Expression` / `Rating` / `Filter` come from the script itself and are corroborated by the live table's item values being ResponseIDs, but a run against the real source before upload would close that gap.
- Regenerating the items CSV needs `Ratings_RawData.csv` and should wait until the corrected table has actually shipped, so the item set matches.

https://claude.ai/code/session_01Wsbf1DRXweudJwB6n9Vczh


ben-domingue:
**Closing this — the diagnosis behind it is wrong. Do not merge.**

I swapped `item` and `id`. But `id <- x$Expression` was already correct, and IRW's established convention for stimulus-rating studies is not a swap at all.

`datastandard.md` line 25: `id` is "the focal unit being measured — typically a person, but sometimes another entity (**e.g., a word in a lexical task**)". In a concreteness norming study the focal unit is the expression, so it belongs on `id`.

Existing tables are consistent on this:

| script | `id` | `item` | `rater` |
|---|---|---|---|
| `KalimahNorms_alzahrani_2025.R` | the Arabic word | `task_name` | participant |
| `emoji_scheffler_2024.R` | the emoji | `trial_name` (dimension) | `submission_id` |
| `famous_melodies.R` | `stimulus` | `condition` | `sub_num` |
| `wine_luckett2021.R` | `Sample.Name` | — | `Subject.Code` |

So the stimulus goes on `id`, the human goes in the `rater` column, and `item` is the rating dimension. The real defect in `data/concretewords.R` is that the rater was put on the **`item` axis** instead of into a `rater` column — not that `id` was wrong.

`KalimahNorms` also shows what to do with the stimulus wording: it keeps it as `cov_arabic_word`, a covariate. The expression should ride along the same way.

Corrected shape:

```r
id    <- x$Expression        # focal unit being measured
item  <- "concreteness"      # the single rating dimension
rater <- x$Participant       # the observer
resp  <- x$Rating
# plus the expression as a cov_ column, per KalimahNorms
```

This also collapses the item-text argument in #1876: with one rating dimension there is one item, and the expression lives on `id`, so there is no per-expression item text to attach. The parking decision stands, but not for the reason recorded.

Superseded by the corrected diagnosis on #1876.

Issue #1831, opened 2026-09-02
Itemtext batch_016: 22 tables across promis1wave1, COACH_Chen_2022 and ecps_sahm_2024


xingyi-zhang:
Next extraction round. 22 tables across **3 studies**, all `AVAILABLE` in the availability audit and all `pending` in `extraction_batches/queue_state.csv`. Follows batch_015 (#1774).

Three clusters rather than 22 unrelated tables, so this is 3 source hunts rather than 22 — and it deliberately exercises the multi-scale-from-one-source case the pipeline has not been through at scale.

## `promis1wave1` — 9 tables

`anger`, `anxiety`, `cesd`, `depression`, `fatigue`, `haq`, `pain`, `physicalfunction`, `social`

Expected to be the high-yield half: PROMIS item banks are publicly and precisely documented, so item text should be explicitly keyed rather than inferred. `cesd` and `haq` are not PROMIS instruments (CES-D and the Health Assessment Questionnaire), so they need separate sourcing — worth not assuming the cluster is homogeneous.

## `COACH_Chen_2022_*` — 5 tables

`ADL`, `CSQ`, `IADL`, `MOS_SSS_C`, `WHOQOL_BREF`

The multi-scale case @ben-domingue flagged: five scales from one study. **Two siblings already have item text** — `coach_chen_2022_hdrs` and `coach_chen_2022_phq9` — so there is in-corpus precedent to match rather than a convention to invent. All five are named published instruments (Activities of Daily Living, Client Satisfaction Questionnaire, Instrumental ADL, MOS Social Support Survey, WHOQOL-BREF), so expect `canonical_instrument` sourcing and the liberal-bar question to arise.

**`WHOQOL_BREF` needs its scoring direction checked against its own data, not assumed.** The corpus already holds the same instrument stored both ways: `altahla_2024_whoqol` keeps the canonical reverse triple raw, while `burkert_2019_whoqol_bref` ships it already reversed. Reverse-coding is a property of the table, not the instrument.

## `ecps_sahm_2024` — 8 tables

`distrust`, `dtl`, `ia`, `identity`, `moral`, `sscd`, `stress`, `vaccine`

Eight short scales from one study. The codes are abbreviations rather than instrument names, so step one is establishing what each measures before any extraction — the `busch_2023_stigma` failure mode, where a table name named the wrong construct.

## Deliverable

[...]

PR #1935, opened 2026-09-04
Item text batch_022: 9 promis1wave1 and 11 ecps_sahm_2024 tables, and a source misattribution (#1831)

xingyi-zhang:
Item text for the two clusters #1831 lists as outstanding, plus the three ecps siblings that were never in its scope. **20 tables, 783 items, 4,050 rows**, of which 3,873 option cells carry a label.

`COACH_Chen_2022_*` is not here: it landed as batch_020 while this was in flight, so those five tables were dropped rather than shipped twice. Findings from the version built here that Ben's supersedes are in the issue thread, including one on `CSQ` q3/q8 that his comment doesn't cover.

| cluster | tables | items | basis | status |
|---|---|---|---|---|
| `promis1wave1_*` | 9 | 660 | `data_labels` | VERIFIED |
| `ecps_sahm_2024_*` | 11 | 123 | `paper_order` | PARTIAL |

## promis1wave1 — 9 of 9

Read from the study's own PROMIS Wave 1 codebook (Appendix A, `doi:10.7910/DVN/0NGAKG`, CC0), which names every code and pairs it with its stem and numbered options. Two codebook quirks are resolved with evidence rather than assumption:

- **Some option sets are 6 wide where live `resp` has 5 levels.** Per-item respondent counts separate two mechanisms. No deficit (PAINBE median n=2,503 against 2,647 for its 5-option siblings) means those respondents are present and level 1 covers both bottom labels, recorded as a `' / '` join. A clear deficit means the extra category is absent: `PAININ21` at n=1,980 against 2,647 drops its "Does not apply to me" escape, and `PFC1`–`PFC5` (n=1,069–2,145 against a 2,535 median, the deficit tracking easiness at Spearman −0.8) drop the top "Very easy".
- **The codebook misprints an option key** on 19 Social Role items, printing 3 twice where the descending ladder needs 2 for "Often"; 17 siblings print 2 for the identical anchor set.

`cesd` and `haq` were initially thought unrecoverable. Both were in the codebook all along — the gap was a parser that required codes to end in a digit, so `HAQ1a` and its 18 siblings never matched. CES-D carries an independent check that doesn't use the codebook: the canonical scale words items 4, 8, 12 and 16 positively, the shipped stems put the positive wording at exactly those numbers, and in the data those four are the only items correlating negatively with the rest (−0.29 to −0.13 against +0.22 to +0.34). `verify_promis1wave1_cesd.R` recomputes it and fails if the separation isn't clean.

`EDANG14` is the only item correlating negatively with its own bank (−0.285), so its labels were checked rather than assumed, and they hold: under the printed ladder 51.1% answered "Never" and 1.7% "Always", where a flipped ladder would claim 51% always found it difficult to express anger. The negative sign is the phrasing, not the coding — documented with a public note telling users not to sum it without reversing.

## ecps_sahm_2024 — 11 tables, and a source correction

**These tables are not the study the dictionary named.** All eleven were attributed to Sahm et al.'s German ERQ-S validation. The data are the **COVIDiSTRESS Global Survey Round II**; its cleaned file is hosted in the ERQ-S OSF project alongside that study's own files, and that project's README points elsewhere for its description, which is how the mix-up arose. Every item code in all eleven tables is a column of that file, which `rederive_ecps.py` re-checks. The dictionary has since been corrected. `_emotion` is the one exception — it genuinely is the ERQ short form, which COVIDiSTRESS II included as a block.

Option labels come from `Copy of survey.pdf`, the administered Qualtrics form, **not** the registration workbooks, whose response designs don't describe what was administered. The form corrected four things:

- the agreement ladder has a **Neutral** midpoint the workbooks omit, so a six-anchor list is really seven options — which is what `resp` 1–7 and 0–6 always were;
- compliance items are past tense as administered;
- information acquisition is a 10-point slider, not 1–7, matching the live 1–10;
- "Not applicable" is a separate column, not a level of `resp`.

The `_0neutral` coding is decoded rather than guessed: the cleaned file carries each item twice, and cross-tabulating `_0neutral` against `_midneutral` is exact — 32 of 33 paired items give `0→4, 1→1, 2→2, 3→3, 4→5, 5→6, 6→7`, so the midpoint is pulled out to 0 and the six named anchors stay at 1–6.

Three tables were held and now ship: `dtl` (the form lists 5 uncertainty items, not 4), `stress` (the form's secondary-stressor order fits the conditional-block counts at 12 of 14 positions; indices 8 and 9 carry 6 and 8 responses against ~9,700 for their block-mates, recorded as a source display defect), and `vaccine` (6 statements against codes numbered 2–6 and 9, so ascending order is matched to printed order — an order inference, hence PARTIAL).

`_emotion`'s order needed deciding, since the form and the workbooks disagree about where one item sits, and that changes which code is which. The data settle it: the ERQ's 4/4 suppression–reappraisal split falls on items 1,2,4,5 against 3,6,7,8 with a within-minus-between gap of **0.268**, the maximum over 2,000 random 4/4 partitions, where the workbook order gives 0.003.

`ecps_source.py` holds the instrument as literals — the Qualtrics print layout interleaves the question and answer flows across pages, so item order can't be read off it. Every string is checked against a source instead: **230 verify against the administered form, 17 against the workbooks, 0 against neither.** Positions the form left unlabelled ship empty rather than invented (177 of 4,050 cells), and every live `resp` level in all eleven tables maps onto a position in its own item's ladder with none left over.

`language` is omitted on all eleven: the survey ran in 47 languages, so no single administered language exists to name. Raised on the issue as a case #1777 doesn't cover.

## Gates

`validate_items.R` 20/20 · `normalize_nulls.R` clean · `irw-validate` exit 0, no findings · `check_provenance.R` clean · `lint_verification.R` clean on 20 rows · `audit_batch.R` 16 PASS + 4 WARN, all four documented findings · **`verify_batch.R` 20/20 against live data.**

Both gates added in #1834 pass. `irw-validate`'s `dup_item_resp` and `resp_ambiguous` are keyed on `(item, resp)`, so the per-item option labels that PROMIS `pain` and the ecps tables legitimately carry are not flagged.

## Notes for review

- Numbering has moved twice under this branch: 016 went to #1709's restart, then 020 and 021 landed while this was in flight. batch_022 was free across main and every remote branch at push time. Nothing about the tables depends on the number — only the directory name, the script paths and the `batch` column.
- `origin/main` is merged in, so this is purely additive: 56 files, 25,831 insertions, 0 deletions.
- Uploading is not in scope; that stays Ben's step.

Closes nothing on its own — #1831 also covers the COACH cluster, already shipped as batch_020.


ben-domingue:
Reviewed this alongside batch_020/021. The extraction side looks solid — purely additive, 22 `verify_*.R` for 20 tables, and the two codebook quirks resolved with evidence rather than assumption is exactly the right instinct. Three things to fix or know about before it lands.

## 1. `queue_state.csv` is not updated — this will cause a duplicate extraction

This PR touches `mapping_verification.csv` and `pending_index_notes.csv` but **not** `itemtext/extraction_batches/queue_state.csv`, so **17 of the 20 tables are still marked `pending`**. Nothing in the pipeline knows they are done.

That is not a distant risk. A round takes the first 12 `pending` rows in file order, and `ecps_sahm_2024_dtl` sits at pending-position 61 — **6 rounds from now** a round will claim it and re-extract it from scratch. `promis1wave1_social` is at position 1069, so the cluster would be re-done piecemeal over the next ~90 rounds. At roughly $60 a round that is real money spent producing files that already exist, and worse, the second extraction would land beside the first with no signal that either is a duplicate.

The fix is to set those 17 rows to `status=done`, `batch=batch_022`, with a timestamp. The other 3 — `ecps_sahm_2024_emotion`, `_support`, `_trust` — are not in `queue_state.csv` at all, which matches this PR's note about siblings that were never in `#1831`'s scope; they need adding as `done` rows or they will never be accounted for.

## 2. Batch-number collision, already resolved on my side

A round I fired this morning also claimed `batch_022` and built its own `itemtables/batch_022/`. Different tables — mine took the next 12 alphabetically (`conspiracy_asd__*` through `cormier_2024_*`), so **nothing here is duplicated work** — but merging both would have fused two unrelated batches into one directory with conflicting `notes.csv`, `provenance.csv`, `verification_merged.csv` and `audit_report.csv`.

**I have renumbered mine to `batch_023`.** This PR keeps `batch_022`; nothing needs changing here. Flagging it because it is the second time this has happened (the COACH tables hit it from the other direction), and the underlying cause is that a batch number is claimed by whoever creates the directory first, with nothing reserving it.

## 3. Expect a conflict with #1922

Both PRs append to `itemtext/mapping_verification.csv` and `itemtext/itemtables/pending_index_notes.csv`. Both are append-only, so whichever merges second gets a trivial conflict — resolve by keeping both sides' rows, not by taking one file wholesale.

## On uploading

Merging is safe whenever — it commits the extractions to the repo, not to the warehouse. **Uploading should wait for a triage pass**, which as far as I can tell has not happened for these 20 yet: re-run the gates live rather than trusting the round's own report, read each `verify_*.R` and judge whether it tests the mapping or the plumbing, then decide stage-or-hold per table. That is the step where batch_020's WHOQOL anchor reversal and batch_021's `conner_2017_curiosity` transcription artifact were caught. The `ecps_sahm_2024_*` cluster is `paper_order` + PARTIAL, which is precisely the combination triage exists to look hard at.

PR #1937, opened 2026-09-04
queue_state: account for the three batch_022 siblings missing from the queue (#1831)

xingyi-zhang:
**Three rows.** Reduced from 20 after batch_022 merged.

The other 17 were set to `done,batch_022` directly on main, so this branch's copies of them differed only by timestamp and were pure conflict — dropped.

These three are the remainder, and they were **never in `queue_state.csv` at all**: `ecps_sahm_2024_emotion`, `_support` and `_trust` are live tables that fell outside #1831's scope and were added to batch_022 because they are the same study, the same administered instrument and the same `_0neutral` coding as the eight ecps tables the issue did list.

With no row, nothing accounts for them, and a regenerated queue would enter them as `pending` and re-extract work that is already on main.

Recorded `done` with `batch=batch_022` and the timestamp main already used for the rest of the batch, per `round_prompt_v1.md` Step 5 ("Table wrote a CSV AND `audit_batch.R` says PASS or WARN -> `status=done`").

Queue after this: **1,404 rows · 1,110 pending · 205 done · 54 excluded · 23 blocked · 12 failed.**

### DIFF
--- a/itemtext/extraction_batches/queue_state.csv
+++ b/itemtext/extraction_batches/queue_state.csv
@@ -1400,3 +1400,6 @@ sun_2025_morality_study2_colleaguerespect,pending,,
 sun_2025_morality_study2_colleaguetrust,pending,,
 wang_onlineriskexp_2025_interpersonalsecurity,pending,,
 wang_onlineriskexp_2025_negativeatt,pending,,
+ecps_sahm_2024_emotion,done,batch_022,2026-09-04T11:00:00-07:00
+ecps_sahm_2024_support,done,batch_022,2026-09-04T11:00:00-07:00
+ecps_sahm_2024_trust,done,batch_022,2026-09-04T11:00:00-07:00


xingyi-zhang:
Superseded by #2227, which carries these three lines.

The finding still stands: `ecps_sahm_2024_emotion`, `_support` and `_trust` are live in `irw_text` with verify scripts in `batch_022` on main, and still have no row in `queue_state.csv`. What went stale is this PR — conflicting since 2026-09-04, and the counts in the body above are now wrong (it says 1,404 rows / 1,110 pending; main is at 1,401 / 111 pending, so the pending figure is off by a factor of ten).

Rather than rebase a three-line PR and rewrite its body, the rows are now in #2227, which is based on current main, already edits `queue_state.csv`, and is mergeable. They are called out under their own heading there so they are not mixed into the batch work.

Issue #1954, opened 2026-09-05
Re-audit already-shipped item text against the 2026-09-04 rights rulings


ben-domingue:
Two rulings on 2026-09-04 narrowed the ECR-R "source licence governs" rule, and **no already-shipped table has been checked against either**.

- **Fee-licensed instruments** (`itemtext_standard.md` § Rights): an enforced licence fee disqualifies an instrument's wording even where IRW's copy came from an openly licensed deposit. Ruled on the TAS-20.
- **No-redistribution clauses**: a quotable clause barring redistribution of the instrument overrides the licence of the deposit the wording was taken from. Ruled on the DSES.

Together these mean **a source deposit's open licence is no longer sufficient on its own** — a rights holder's own restriction outranks it.

**The problem.** Everything shipped before 2026-09-04 went out under the old rule, which asked only about the deposit. Nobody has since asked, for those tables, whether the *instrument's* rights holder states a fee or a redistribution bar. Some will not survive.

**`wording_rights` cannot find them.** `CV_OASIS_ODSIS_PPE_Novak_2020_DSES` was the only table in the entire corpus carrying that column, and it has now been withdrawn. There is no flag to filter on; the sweep has to read `provenance.csv` and re-check each instrument's rights holder.

**Suggested shape.** Work from `provenance.csv` across all batches (~347 rows, ~270 tables). Tables whose `text_source` is `canonical_instrument` are the highest-risk group — those took wording from the published instrument rather than the study's own materials, which is exactly where a publisher's terms attach. `study_materials` tables are lower risk but not zero, since the instrument's terms can still bar redistribution of wording a study reproduced.

Scale is real: this is a per-instrument rights check, not a per-table one, so it should collapse to far fewer than 270 lookups once instruments are grouped.

Context: #1709, and the round log entries for 2026-09-04.

Issue #1945, opened 2026-09-04
Itemtext batch_201-205: 60 tables from the queue tail, in reverse file order


xingyi-zhang:
Claiming **batch_201 through batch_205** — 60 tables, taken from the **tail** of `extraction_batches/queue_state.csv` in reverse file order. Follows batch_022 (#1831). Numbering and direction agreed with @ben-domingue.

## Why claim a range, and why from the tail

Hand-built batches and the round runner were drawing from the same two shared resources, and collided repeatedly. batch_022 was renumbered twice before shipping — 016 went to #1709's restart, then 020 and 021 landed while it was in flight — and its 17 queued tables sat `pending` after merging, which is the same gap that produced the duplicated `COACH_Chen_2022_*` extraction.

Both resources are now partitioned rather than shared:

**Numbering.** `run_round.sh` takes the next sequential `batch_NNN`. With 1,110 pending at 12/round the runner's sequence tops out near `batch_116`, so a hand-built range at 201+ cannot intersect it. Claiming five at once also means the range is declared once rather than renegotiated per batch.

**Order.** A round takes the first 12 rows with `status=="pending"` in **file order**. These 60 are file positions **1051–1110 of 1110** — the far end, roughly 92 rounds of clearance. Note file order is *not* alphabetical (Spearman 0.655 against alphabetical rank), so reverse-*alphabetical* would only have reached position ~989; reverse *file* order is the one that maximises the gap.

Rows will be set to `status=done, batch=batch_NNN` with a timestamp as each batch lands, per `round_prompt_v1.md` Step 5 — not left `pending`, and not `excluded`.

## The 60 tables

<details><summary><b>batch_201</b> — PLOS + OSF, the most promising of the five</summary>

`wang_onlineriskexp_2025_negativeatt`, `wang_onlineriskexp_2025_interpersonalsecurity`, `sun_2025_morality_study2_colleaguetrust`, `sun_2025_morality_study2_colleaguerespect`, `sun_2025_morality_study1_fairnessMCQ`, `sun_2025_morality_study1_fairnessHEXACO`, `socialstereotype_hughes_2025_judgement`, `singh_2025_identity_pba`, `shu_2025_translation_add`, `pezzuti_2025_coolpeople_main_nfc`, `paampsmartsud_saba_2023_ders`, `oxfordcovid_xue_2024_at`
</details>

<details><summary><b>batch_202</b> — one 8-table OSF deposit, then the gilbert_meta family starts</summary>

[...]

PR #1993, opened 2026-09-05
Restore complete Su 2024 Xinxiang conversion outputs

sinew-07:
Running the committed Xinxiang conversion from a fresh working directory produces only PHQ-9 because the calls for GAD-7, ISI and PSS-14 are missing. This restores those calls and the four-scale QC summary, adds verified source documentation, and writes the QC artifact to `output/qc/su_2024_quality_check.csv`. The four response CSVs remain directly under `output/`.

This repairs the provenance of four existing IRW tables. It does not add or re-upload data.

### Source and RT units

The script now cites [Su et al. (2024), Scientific Data 11, 1046](https://doi.org/10.1038/s41597-024-03888-8) and the specific [Zenodo data release, DOI 10.5281/zenodo.10423537](https://zenodo.org/records/10423537). The deposit explicitly licenses the data under **CC BY 4.0**. All five local input files match the sizes and MD5 checksums published by that Zenodo record.

**The source RT fields are already in seconds.** The paper's Table 1 (PDF page 3) defines `timeX` for each of the four CSVs as “Time taken (in seconds) for the participant to answer the Xth question.” Methods → Data generation process also states two-decimal precision. The converter preserves these values; dividing by 1,000 would change the documented units incorrectly.

The following medians pool all participant–item RT records within each scale, rather than participant total completion times or averages of item medians. They were independently recomputed from the raw exports and agree exactly with the rebuilt outputs and QC CSV.

| Table | Rows | Participants | Items | Observed response range | Median RT (seconds) |
|---|---:|---:|---:|---|---:|
| `su_2024_gad7` | 170,044 | 24,292 | 7 | 0–3 | 2.20 |
| `su_2024_isi` | 170,044 | 24,292 | 7 | 0–4 | 2.65 |
| `su_2024_phq9` | 218,628 | 24,292 | 9 | 0–3 | 3.00 |
| `su_2024_pss14` | 340,088 | 24,292 | 14 | 1–5 | 3.22 |

One coding clarification: the deposited PSS question fields contain **1–5**, which this conversion preserves. The paper's supplementary table describes a 0–56 total and reverse-scoring examples, but does not provide a mapping for the released CSV codes. This repair does not subtract 1, reverse-score items or infer a total-score transformation.

### Validation

- Re-ran the revised script with `Rscript --vanilla` in a new directory containing copies of the five source files. Exactly four response CSVs are written at the top level, with the QC CSV under `output/qc/`.
- All four response files are byte-for-byte identical to the existing local outputs. The relocated QC file also has identical content; only its destination changes. The earlier baseline reproduction produced PHQ-9 alone.
- Independently compared all **898,804 participant–item records**, including responses, RTs and five demographic covariates (**6,291,628 fields**): zero differences; participant sets, item counts, complete response distributions and missingness are preserved. Original row ordering is unchanged.
- The 57 `irw_validate` tests, 58 `red_up` tests, R validator smoke test and modified-script contract check pass locally. All 979 files in the original workspace retain their recorded hashes and modification times.

Full-data reproduction used R 4.5.3. Local Python tests used Python 3.12 for `irw_validate` and the existing Python 3.14 environment for `red_up`; GitHub CI uses the repository's Python 3.12 workflow. Source/participant records and generated tables are not included in this PR. Earlier root-level QC artifacts are not automatically deleted from existing working directories; the output-layout check above uses a fresh directory.

The PR remains **Draft** for review.

### DIFF
--- a/data/su_2024_xinxiang_assessments.R
+++ b/data/su_2024_xinxiang_assessments.R
@@ -294,3 +294,97 @@ convert_scale_to_irw <- function(
   
   return(long_data)
 }
+# GAD-7 item map
+gad7_item_map <- tibble(
+  item = paste0("GAD_", 1:7),
+  resp_col = paste0("question", 1:7),
+  rt_col = paste0("time", 1:7)
+)
+
+gad7_item_map
+# 转换并导出 GAD-7
+gad7_long <- convert_scale_to_irw(
+  input_file = "gad7.csv",
+  item_map = gad7_item_map,
+  output_file = "su_2024_gad7.csv"
+)
+# ISI item map
+isi_item_map <- tibble(
+  item = paste0("ISI_", 1:7),
+  resp_col = paste0("question", 1:7),
+  rt_col = paste0("time", 1:7)
+)
+
+isi_item_map
+
+# 转换并导出 ISI
+isi_long <- convert_scale_to_irw(
+  input_file = "isi.csv",
+  item_map = isi_item_map,
+  output_file = "su_2024_isi.csv"
+)
+# PSS-14 item map
+pss_item_map <- tibble(
+  item = paste0("PSS_", 1:14),
+  resp_col = paste0("question", 1:14),
+  rt_col = paste0("time", 1:14)
+)
+
+pss_item_map
+
+# 转换并导出 PSS-14
+pss_long <- convert_scale_to_irw(
+  input_file = "pss.csv",
+  item_map = pss_item_map,
+  output_file = "su_2024_pss14.csv"
+)
+# =========================================================
+# 四个量表的总质量检查
+# =========================================================
+
+all_scales <- bind_rows(
+  phq9_long |> mutate(dataset = "PHQ-9"),
+  gad7_long |> mutate(dataset = "GAD-7"),
+  isi_long  |> mutate(dataset = "ISI"),
+  pss_long  |> mutate(dataset = "PSS-14")
+)
+
+overall_qc <- all_scales |>
+  group_by(dataset) |>
+  summarise(
+    rows = n(),
+    ids = n_distinct(id),
+    items = n_distinct(item),
+    min_resp = min(resp, na.rm = TRUE),
+    max_resp = max(resp, na.rm = TRUE),
+    missing_resp = sum(is.na(resp)),
+    missing_rt = sum(is.na(rt)),
+    zero_rt = sum(rt == 0, na.rm = TRUE),
+    negative_rt = sum(rt < 0, na.rm = TRUE),
+    min_rt = min(rt, na.rm = TRUE),
+    median_rt = median(rt, na.rm = TRUE),
+    mean_rt = mean(rt, na.rm = TRUE),
+    max_rt = max(rt, na.rm = TRUE),
+    .groups = "drop"
+  )
+
+overall_qc
+# 检查输出文件
+output_files <- c(
+  "su_2024_phq9.csv",
+  "su_2024_gad7.csv",
+  "su_2024_isi.csv",
+  "su_2024_pss14.csv"
+)
+
+file_check <- tibble(
+  file = output_files,
+  exists = file.exists(file.path(output_dir, output_files)),
+  size_bytes = file.info(file.path(output_dir, output_files))$size
+)
+
+file_check
+write_csv(
+  overall_qc,
+  file.path(output_dir, "su_2024_quality_check.csv")
+)
--- a/data/su_2024_xinxiang_assessments.R
+++ b/data/su_2024_xinxiang_assessments.R
@@ -1,3 +1,15 @@
+# Su et al. (2024), Temporal dynamics in psychological assessments:
+# a novel dataset with scales and response times. Scientific Data 11, 1046.
+# Paper: https://doi.org/10.1038/s41597-024-03888-8
+# Data: https://doi.org/10.5281/zenodo.10423537
+# Source: https://zenodo.org/records/10423537
+# License: CC BY 4.0 (data deposit), https://creativecommons.org/licenses/by/4.0/
+# Inputs: demographic.csv, phq9.csv, gad7.csv, isi.csv, pss.csv in raw/.
+# RT units: seconds, recorded to two decimal places (paper, Methods,
+# "Data generation process"). Preserve time1..N without a /1000 conversion.
+# PSS-14: retain the 1-5 item codes stored in the deposited pss.csv.
+# This restores reproducibility of existing IRW tables; no new data are added.
+
 library(tidyverse)
 library(janitor)
 library(readr)
@@ -384,7 +396,10 @@ file_check <- tibble(
 )
 
 file_check
+# QC summaries are not IRW response tables; keep them out of output/*.csv.
+qc_dir <- file.path(output_dir, "qc")
+dir.create(qc_dir, showWarnings = FALSE)
 write_csv(
   overall_qc,
-  file.path(output_dir, "su_2024_quality_check.csv")
+  file.path(qc_dir, "su_2024_quality_check.csv")
 )


ben-domingue:
Thanks — I read through this against the current script and the metadata.

To be clear about what this is for anyone else looking: it's not new data. All four tables are already published in IRW:

| table | rows | ids | items | categories |
|---|---|---|---|---|
| `su_2024_gad7` | 170,044 | 24,292 | 7 | 4 |
| `su_2024_isi` | 170,044 | 24,292 | 7 | 5 |
| `su_2024_phq9` | 218,628 | 24,292 | 9 | 4 |
| `su_2024_pss14` | 340,088 | 24,292 | 14 | 5 |

So this is a provenance repair. `su_2024_xinxiang_assessments.R` currently ends at the closing brace of `convert_scale_to_irw()` — the function is defined and never called for GAD-7, ISI, or PSS — so a fresh run reproduces one of the four tables that are live. That's a real defect and worth fixing; nothing needs re-uploading.

The arithmetic checks out independently: 24,292 × (7 + 7 + 9 + 14) = 898,804, exactly the record count you compared. Category counts match the expected 0–3 (PHQ-9, GAD-7) and 0–4 (ISI, PSS-14) ranges.

Things I'd like addressed:

1. **No source header.** `datastandard.md` requires a comment block at the top with the source URL and DOI. This script has neither — no paper, no data link, no licence line — and most other `data/*.R` files carry `# Paper:` / `# Data:`. Since you're already in the file, please add it. These headers are what our retrieval work reads, so it matters more than it looks.

2. **`rt` units are unverified, and this is the one I actually want answered.** The standard is seconds only, convert from milliseconds if needed. The script passes `time1..N` straight through with no divisor. Your added QC computes `min_rt` / `median_rt` / `max_rt` but the numbers aren't in the PR description — could you paste the per-scale medians? If these are milliseconds, then four published tables are wrong, which is a considerably bigger finding than the one this PR fixes.

3. **`su_2024_quality_check.csv` lands in `output/`** alongside the four IRW files. It has no `id`/`item`/`resp` so validation would reject it, but a QC artifact sitting in the output directory is asking for a bad glob some day. `output/qc/` or the script directory would be safer.

Two things I'm not asking you to change:

- Row order differs between PHQ-9 (sorted by `item_number`, numeric) and the rest (sorted by `item`, so `PSS_10` precedes `PSS_2`). Cosmetic, and already baked into what's published.
- I can't independently confirm the byte-identity claim or the 898,804-record comparison from the PR alone, since the raw data isn't included — that's expected. The row counts matching metadata is a good sign.

Happy to merge once (1) is in and (2) has an answer.

ben-domingue:
Merged.

Recording the answer to (2) here so it is findable outside the PR description: **the `rt` values in the four `su_2024_*` tables are seconds, and the published tables are correct.** The per-scale medians are GAD-7 2.20, ISI 2.65, PHQ-9 3.00, PSS-14 3.22. Those settle it independently of the paper — as milliseconds they would be 2-3ms per Likert item. No `/1000` conversion should be applied here, and the script header now says so inline.

(1) and (3) are in: source/DOI/licence header (CC BY 4.0), and the QC artifact moved to `output/qc/`.

Not fixed, as agreed: row order still differs between PHQ-9 (numeric `item_number`) and the other three (string `item`, so `PSS_10` precedes `PSS_2`), matching what is already published.

ben-domingue:
Follow-up, and the more important half of this: **the script is now fixed, but all four of these tables still have no dictionary entry.** Merging this repaired reproducibility; it did not give the tables provenance.

Checked against the live core dictionary sheet and the committed pipeline outputs:

| where | `su_2024_gad7` / `isi` / `phq9` / `pss14` |
|---|---|
| core dictionary sheet | **0 rows** |
| `metadata/biblio.csv` | **0 rows** — no DOI, no data URL, no `Derived_License`, no description |
| `tags/` | **no occurrences**, and not in `dict_minus_tags_queue.txt` (513 entries) |
| `metadata/metadata.csv` | present, counts correct |

`metadata.csv` is the trap here. It is computed from the live tables, so these four look healthy in the file most coverage checks read, while carrying no source or licence anywhere.

Worth flagging how they got this way, because it can recur: `metadata/pipeline_logs/NEXT_RUN_NOTES.md` records 10 "stale biblio-only rows" being dropped, `su_2024_*` among them. They were not stale — the tables are live. The orphan (a biblio row with no backing dictionary row) got resolved by deleting the provenance rather than creating the missing dictionary row. `.claude/skills/irw-site-update/TODO.md` had separately flagged `su_2024_isi`/`phq9`/`pss14` as "no dictionary entry at all… not yet investigated". So this is the second independent sighting.

The good news is that this PR supplies exactly the fields that are missing, verified against the deposit:

- Paper DOI `10.1038/s41597-024-03888-8`
- Data DOI `10.5281/zenodo.10423537` (https://zenodo.org/records/10423537)
- **Licence CC BY 4.0** — no fee, no ND, no stated use restriction, so it clears the tightened bar from #1945

On instrument rights, which is the obvious worry with four named clinical scales: it does not apply. These ship item *labels* (`PHQ_1`), not wording — no `option_text` — and none of the four are in the item-text corpus. The deposit licence is the only licence in play, so this is not in scope for #1954.

**What's needed:** four rows in the core dictionary sheet. That is a Sheet edit, not a code change — once they exist, `02_biblio.R` populates biblio and the tagging queue picks them up without further work.

sinew-07:
> Follow-up, and the more important half of this: **the script is now fixed, but all four of these tables still have no dictionary entry.** Merging this repaired reproducibility; it did not give the tables provenance.
> 
> Checked against the live core dictionary sheet and the committed pipeline outputs:
> 
> where	`su_2024_gad7` / `isi` / `phq9` / `pss14`
> core dictionary sheet	**0 rows**
> `metadata/biblio.csv`	**0 rows** — no DOI, no data URL, no `Derived_License`, no description
> `tags/`	**no occurrences**, and not in `dict_minus_tags_queue.txt` (513 entries)
> `metadata/metadata.csv`	present, counts correct
> `metadata.csv` is the trap here. It is computed from the live tables, so these four look healthy in the file most coverage checks read, while carrying no source or licence anywhere.
> 
> Worth flagging how they got this way, because it can recur: `metadata/pipeline_logs/NEXT_RUN_NOTES.md` records 10 "stale biblio-only rows" being dropped, `su_2024_*` among them. They were not stale — the tables are live. The orphan (a biblio row with no backing dictionary row) got resolved by deleting the provenance rather than creating the missing dictionary row. `.claude/skills/irw-site-update/TODO.md` had separately flagged `su_2024_isi`/`phq9`/`pss14` as "no dictionary entry at all… not yet investigated". So this is the second independent sighting.
> 
> The good news is that this PR supplies exactly the fields that are missing, verified against the deposit:
> 
> * Paper DOI `10.1038/s41597-024-03888-8`
> * Data DOI `10.5281/zenodo.10423537` (https://zenodo.org/records/10423537)
> * **Licence CC BY 4.0** — no fee, no ND, no stated use restriction, so it clears the tightened bar from [Itemtext batch_201-205: 60 tables from the queue tail, in reverse file order #1945](https://github.com/ben-domingue/irw/issues/1945)
> 
> On instrument rights, which is the obvious worry with four named clinical scales: it does not apply. These ship item _labels_ (`PHQ_1`), not wording — no `option_text` — and none of the four are in the item-text corpus. The deposit licence is the only licence in play, so this is not in scope for #1954.
> 
> **What's needed:** four rows in the core dictionary sheet. That is a Sheet edit, not a code change — once they exist, `02_biblio.R` populates biblio and the tagging queue picks them up without further work.

I’ve added and checked the four Su entries in the core dictionary, including the source DOIs, CC BY 4.0 licence and scale descriptions. The response data are unchanged.

ben-domingue:
Checked against the live core dictionary sheet — the four rows are there (contributor `sinew`, dated 9/6/2026), and they're complete:

| field | value |
|---|---|
| `DOI (for paper)` | `10.1038/s41597-024-03888-8` |
| `URL (for data)` | `https://zenodo.org/records/10423537` |
| `Original License` / `Derived License` | CC BY 4.0 (both populated) |
| `Public Reshare?` | Public |

The descriptions agree with the live tables rather than with the paper — 24,292 participants across all four, and 7 / 7 / 9 / 14 items, exactly what `metadata.csv` reports. The `pss14` note recording that released item codes 1–5 are preserved without reverse scoring is the right call to have documented.

`metadata/biblio.csv` and `tags/` are still empty for these four. That is expected and needs no further edit — both are computed downstream of the dictionary, and they fill on the next `metadata-pipeline` run. That run's PR is the confirmation worth reading: if the biblio rows don't appear with `Derived_License = CC BY 4.0`, the orphan survived in a different form.

One thing that came out of checking the rest of that batch. Of the 10 "stale biblio-only rows" dropped, the other six — `adhd_silk_2019_*`, `alcoholhealthwarninglabel_brennan_2022_positive_arousal`, `racialsocialnormsbrazilianstudents_portella_2022` — have no `metadata.csv` row at all, so those tables aren't live and deleting them was correct. The trap caught only `su_2024_*`, which is now closed. `.claude/skills/irw-site-update/TODO.md` had all six of its "no dictionary entry at all" rows outstanding; #1999 closes that item out and records why the live-table check is the one that matters before dropping a biblio row.

Nothing outstanding here.

Issue #1754, opened 2026-08-31
nguyen_2026_{barthel,gad7,isi,mspss,pic}: 5 live tables with no dictionary entry and no identifiable contributor


ben-domingue:
Five tables are live in Redivis (`item_response_warehouse_4`) with no data dictionary entry, no biblio entry, no tags, and no identifiable contributor. Surfaced by the weekly metadata audit on 2026-08-31.

## The tables

| table | n_responses | n_participants | n_items | n_categories |
|---|---|---|---|---|
| `nguyen_2026_barthel` | 3,940 | 394 | 10 | 4 |
| `nguyen_2026_gad7` | 2,758 | 394 | 7 | 4 |
| `nguyen_2026_isi` | 2,758 | 394 | 7 | 5 |
| `nguyen_2026_mspss` | 4,728 | 394 | 12 | 7 |
| `nguyen_2026_pic` | 3,546 | 394 | 9 | 5 |

Identical N=394 across all five, so this is one sample measured on five instruments — Barthel Index (ADL), GAD-7, Insomnia Severity Index, MSPSS (perceived social support), and PIC.

## Why we can't attribute them

- **No processing script in `data/`.** The three `nguyen_2026_*` scripts there (`nguyen_2026_learner_autonomy.py`, `nguyen_2026_online_learning_misfit.py`, `nguyen_2026_sdt_online_learning.py`) produce the `autonomy_*` / `misfit_*` / `sdt_*` families — 18 tables, all properly documented. None of them produce these five.
- **Not from `automated_finding`.** `BATCH_LOG.md` mentions only those same three scripts under `nguyen`.
- **No GitHub issue** references them.
- **Not a naming mismatch.** Checked the full 4,301-row dictionary for near-matches on `barthel`, `gad7`, `mspss`, `isi` and `insomnia` — every hit belongs to an unrelated author (`amarilla_2020_barthel`, `laksmita_2020_mspss`, `isi_insomnia_wang2025`, and a dozen unrelated `*_gad7` tables). These are genuinely missing rows, not misfiled ones.

Despite the shared `nguyen_2026` prefix, the instrument set (clinical/functional scales) looks unrelated to the online-learning Nguyen datasets, and the prefix collision may itself be worth resolving — two different studies currently share it.

## What's needed

1. Identify who uploaded these and which source study they come from.
2. Add the 5 dictionary rows (which will generate the biblio entries) and tag them.
[...]

Issue #1690, opened 2026-08-27
465 dictionary rows hold a dataset DOI in the `DOI (for paper)` column


ben-domingue:
Surfaced by the table-name audit (#1686) but distinct from it: the audit could not have
found the naming errors without first normalising this column, and it is what generates
most of that audit's false positives.

## The problem

**465 of 1,945 automated-pipeline rows hold a *dataset* DOI in the `DOI (for paper)`
column** — figshare (`10.6084`), Mendeley Data (`10.17632`), Harvard Dataverse
(`10.7910`), OSF (`10.17605`), Zenodo (`10.5281`), Dryad (`10.5061`). These are the
deposit record, not the paper.

That matters beyond tidiness:

- **The year is wrong.** A deposit year is not a publication year, which is what produced
  ~55 spurious `year_mismatch` flags in the audit. The table names in those cases are
  most likely correct and the DOI column is the thing that's off.
- **It is the wrong citation target.** Anyone resolving `DOI (for paper)` expecting the
  article gets a data deposit.
- **Any future automated check inherits the error.** The naming gate proposed in #1686
  resolves this column to get an author list; where the column holds a dataset DOI it
  gets the depositor rather than the paper's authors, which for Dataverse in particular
  is frequently a different person (`cavojova_2017_cfc` and `cosenza_2015_cfc` both look
  fabricated for exactly this reason, and both are correctly named).

## Other malformed values in the same column

| pattern | rows | example |
|---|---|---|
| full URL rather than a bare DOI | 33 | `https://doi.org/10.7910/DVN/ZDNSFJ` |
| prefixed with the literal text `data doi: ` | 22 | `data doi: 10.6084/m9.figshare.26820745.v4` |
| PLOS supplement suffix `.sNNN` | 10 | `10.1371/journal.pone.0340806.s001` |
| free text | 4 | `not yet published` |
| empty | 30 | |
| multiple DOIs / notes crammed into one cell | 2 | `condon_2024_sapa_personality` holds seven DOIs |

## Suggested work

1. **Normalise the mechanical cases** — strip `data doi: ` prefixes, unwrap
   `https://doi.org/` URLs, drop PLOS `.sNNN` suffixes. Cheap and unambiguous.
2. **For the 465 dataset DOIs, decide the schema question first.** Two options, and it is
   a real choice rather than a cleanup: either resolve the linked publication and put the
   article DOI in `DOI (for paper)`, keeping the dataset DOI in `URL (for data)`; or add a
[...]

PR #1994, opened 2026-09-05
Restore Nguyen 2026 surgical-study provenance and metadata

sinew-07:
The five published Nguyen surgical-study tables lacked a recoverable conversion script and identifiable source metadata. This PR reconstructs the converter from [Harvard Dataverse, DOI 10.7910/DVN/X2C2PL, version 1.0](https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/X2C2PL&version=1.0), pins the workbook checksum, and repairs the resolver's malformed BibTeX author using the native deposit metadata.

The converter preserves all five table names, 394 participant IDs, 45 items, 17,730 responses, row order and CSV layout. It refuses an incorrect source checksum, existing target files or failed validation before writing any response CSV. The original processing script and original uploader identity were not recovered; Sinew is credited for this reconstruction and metadata submission.

The exact-DOI citation correction covers both fresh retrieval and cached citations. It follows the current pipeline's paper-first/data-DOI fallback, including cached rows after the DOI-column split. Unrelated citations remain unchanged. A regression runs in CI.

### Validation

- Re-ran the unchanged converter after merging current main. All five outputs are byte-identical to the **existing local CSVs**, corresponding ZIP members and prior reconstructed outputs. This is not an independent byte comparison against all five currently published tables.
- Independently matched all 17,730 participant-item responses to the source workbook, including complete distributions, participant/item sets, missingness and row order. Original local files retained their hashes and modification times.
- Twelve rejection cases verified that failures stop before any output CSV is written, including a validation failure on the fifth scale.
- Python validator/uploader and related repository tests, R validator/dictionary/tag tests, the added BibTeX tests and the converter contract check pass locally. The initial two R dependency failures were resolved by using the existing project R library; their reruns pass.

### Current handoff

The dictionary and tag CSVs preserve the original proposals. A fresh check on 9 September confirms that the five rows already exist exactly once in both authoritative Sheets, with tag Rater `sinew`; the reviewed classifications and public source/CC0/description/tag entries are present. No repeat import is needed.

The remaining publication step is the BibTeX author correction: the committed bibliography and public catalogue still contain `ich Nguyen, Bich`, while the deposit names **Van Bich Nguyen**. After merge, the normal metadata regeneration and review/publication workflow should propagate this correction and verify it in the released bibliography and site. The current Redivis metadata release and unpublished drafts were not independently queried in this check.

Keep the existing public table identifiers; any coordinated rename can be considered separately. No raw workbook, response CSV or generated metadata is committed, and no response re-upload is required. This PR references #1754 without automatically closing it; the final citation publication should be verified first. See the [provenance and handoff notes](https://github.com/sinew-07/irw/blob/sinew/nguyen_2026_provenance/processing_notes/metadata_repairs/1754_nguyen_2026/README.md).


ben-domingue:
@sinew-07 — checking in on this. It's been sitting since 5 Sep with no review, which is our fault rather than yours: **both checks are green** (`contract` and `tests`), and reading it back, the work looks done.

It also answers the part of #1754 we thought was hardest. The issue said we couldn't identify the source study at all; you've traced it to Harvard Dataverse [10.7910/DVN/X2C2PL](https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/X2C2PL&version=1.0), reconstructed the converter, pinned the source checksum, and reproduced all 17,730 responses byte-identically against the published tables. That's the answer to question 1, and the `ich Nguyen, Bich` resolver bug is a genuinely useful catch on its own.

**Two questions, so we can move it:**

1. Is it still a draft because *you* have work left, or only because of the handoff you describe — a human importing the five dictionary rows into the sheet and filling the tag `Rater`? If it's the latter, that's ours to do and you can mark it ready for review.
2. `nguyen_2026_` is currently shared by two unrelated studies (these five clinical scales, and the online-learning `autonomy_*`/`misfit_*`/`sdt_*` families). #1754 asks whether the prefix should be disambiguated. Do you have a view? Renaming live tables is our call, not a blocker on this PR — we'd just rather decide it with the person who has read both sources.

One thing worth saying plainly: we have not independently re-verified the byte-identical claim or the checksum pin. That's a review task on our side, not a doubt about the work.

Thanks for this — the provenance notes distinguishing verified source facts from unverified administration details are exactly the right shape.

sinew-07:
Thanks, Ben — marked ready for review. Current main is incorporated and the conflicts are resolved. The merge also needed a small regression-backed fix so the cached citation correction still works after a deposit DOI moves to the new data-DOI column. Both CI checks pass.

The five dictionary and Tags rows are already imported, with Rater `sinew`, and their source, CC0, descriptions and tags are visible in the public catalogue. I updated the handoff notes accordingly. The remaining publication step is to regenerate and publish the corrected BibTeX author after merge; the current generated fields still say `ich Nguyen, Bich`.

On naming, I would retain the five existing public identifiers and distinguish the clinical deposit from the online-learning deposits through their source DOIs and study descriptions. The native records identify Van Bich Nguyen for the clinical deposit and Xiem Nguyen for the three education deposits. Any rename can be a separate coordinated migration.

One precision point on the validation claim: the byte comparison was against the existing **local** CSVs and ZIP, not an independent byte comparison against all five current published tables. A fresh run still reproduces those local files exactly; the source-to-response comparison covers all 17,730 records.

ben-domingue (approved):
Thanks — and thanks especially for the correction on the validation claim. You were right to narrow it: what I wrote said "byte-identically against the published tables," and your check hadn't done that. It matters here more than it might elsewhere, because our local CSV archive is an old snapshot and a meaningful share of it disagrees with what Redivis actually serves. So "matches the local files" wouldn't have been enough.

We've now run the comparison that was missing, against the live tables.

**Checksum pin.** Fresh pull of Dataverse file 14113156 gives MD5 `304a3442b9e3b0700410e65612f89602` and SHA-256 `d6642499ad8fea23787cd84372be70d6a7847b2dc893587b9ef57a0a2932df68`. Both match what the converter pins.

**Converter.** Ran `data/nguyen_2026_gyn_surgery.py` from this branch against that workbook. Clean, 17,730 rows over five tables, and the only output is the four advisory warnings your README already documents (`resp_scale_mixed` on mspss/barthel, `imputed_values` on gad7/barthel). Nothing unexpected.

**Against the published tables.** All five resolve in `item_response_warehouse_4`. Pulled every row and compared:

| table | published | reproduced | value diffs |
| --- | ---: | ---: | ---: |
| `nguyen_2026_barthel` | 3,940 | 3,940 | 0 |
| `nguyen_2026_gad7` | 2,758 | 2,758 | 0 |
| `nguyen_2026_isi` | 2,758 | 2,758 | 0 |
| `nguyen_2026_mspss` | 4,728 | 4,728 | 0 |
| `nguyen_2026_pic` | 3,546 | 3,546 | 0 |

Identical on the full `(id, item)` key set and on every `resp` value — 17,730 for 17,730, 394 participants, 45 items. No nulls, no literal `NA`, and response ranges match the bounds you document per instrument.

One thing worth recording so nobody reopens it later: raw row order from Redivis does *not* match the CSV order. That is storage scan order, not stored order — the rows come back clustered by `resp`. Sorted on `id,item` the five tables are identical. It is not a discrepancy.

So the reconstruction is confirmed against what we actually publish, which is a stronger statement than either of us had made.

**On the citation override.** Reviewed separately. Two things I checked rather than assumed: the relative `source("bibtex_overrides.R")` adds no new constraint, since `02_biblio.R` already sources three siblings the same way and writes `biblio.csv` relatively; and the override actually fires on the real cached row — live `metadata/biblio.csv` carries this DOI in `DOI__for_paper_` with `DOI__for_data_` a literal `NA`, which is exactly the paper-first path `apply_bibtex_overrides` takes. So the correction lands on the next pipeline run without a cache bust.

The test file is the strongest part of this PR. Parsing `fetch_bibtex_from_doi` out of `02_biblio.R` and evaluating it against a mocked `GET` — asserting zero HTTP calls on an overridden DOI and one otherwise — tests the production function rather than a copy of it. The `#1690` split coverage including the literal-`NA` paper column is our recurring foot-gun, and you handled it unprompted.

One thing to record for whoever adds the second entry: the override is permanent and has no recheck. If the resolver's name parsing is fixed upstream we will keep forcing our copy indefinitely and nobody will notice. The `reason` column is most of the mitigation. Not a change request — a note for when this file has ten rows instead of one.

**On the other two points.** Agreed on naming: keep the five identifiers, disambiguate through source DOI and study description. The Van Bich Nguyen / Xiem Nguyen distinction is a useful detail we did not have, and it goes in the notes rather than in a rename. Any migration is a separate coordinated decision.

The dictionary and Tags import being already done on your side is appreciated, though you should not have had to do it — that was ours.

Remaining after merge is the BibTeX author: the generated fields still read `ich Nguyen, Bich` and need a metadata regeneration to reach the committed bibliography and the public catalogue. Keeping #1754 open until that is visible in the released bibliography, per the `Refs` rather than `Closes` you used.

Issue #1996, opened 2026-09-06
data/oxfordcovid_xue_2024.r: the _at table's integer filter deletes 23% of the eight hours-per-day items


xingyi-zhang:
Found during batch_201 item-text extraction (#1945) and confirmed by replicating the script against the raw deposit.

## The filter

`data/oxfordcovid_xue_2024.r:204`:

```r
at_df <- at_df %>%
  filter(!is.na(resp) & resp == floor(resp) & resp <= 5)
```

This is the only one of the script's sixteen `filter(!is.na(resp))` calls that carries extra conditions, and `oxfordcovid_xue_2024_at` is the only one of its seventeen tables affected. Unlike #1971 this does **not** read like a copy-paste slip — there is real junk in these columns (see below), so it looks like a deliberate attempt to clean it. The problem is the rule it chose.

## Why it misfires

The `_at` table pools two different response formats, which the deposit's own dictionary distinguishes:

| `Field_Type` | items | what it is |
|---|---|---|
| `radio` | 36 | `1, Never \| 2, Rarely \| 3, Sometimes \| 4, Often \| 5, Always` |
| `text` | 8 | a free numeric quantity — **hours per day** |

For the eight `text` items the dictionary lists **no** response options at all, and the section prompt explicitly invites decimals:

> Please report the average number of hours per day (you can add a decimal point if you want, e.g. 2.5 if you did 2 and a half hours) you typically engaged in the following activities during the last week:

So `resp == floor(resp) & resp <= 5` deletes exactly the answers the question asked for. It is a no-op on the 36 `radio` items, which are already integers ≤ 5.

## What it deletes

Replicating the script's wave filter and column selection against `ARC_raw_data.csv` reproduces the live table's row count exactly (179,937), so these numbers line up with what is published:

| | responses |
|---|---|
| non-missing in raw, after the wave filter | 192,922 |
| surviving line 204 | 179,937 |
| **deleted** | **12,985 (6.7% of the table)** |

All 12,985 come from the eight `text` items; the 36 `radio` items lose nothing. Concentrated on those eight, the loss is **23.0%**:

| item | raw n | kept | deleted | % lost | raw max |
|---|---|---|---|---|---|
| `activity_exercise_inside` | 7,033 | 4,824 | 2,209 | 31.4% | 60 |
| `activity_exercise_outside` | 7,031 | 4,699 | 2,332 | 33.2% | 333333 |
[...]

Issue #1891, opened 2026-09-04
cdm_timss07: rule on IEA redistribution terms before shipping item text (and how the corpus expresses a mixed-licence table)


ben-domingue:
`cdm_timss07`'s item text was extracted cleanly in batch_018 (Step 5b VERIFIED, all gates green) but is **held from upload** pending a rights decision. This is not a question about the extraction — it is about whether IRW may redistribute IEA's item wording.

## The clause

IEA's TIMSS 2007 released-items notice reads:

> Released items … are for **non-commercial, educational, and research purposes** only.

> **Commercial exploitation, distribution, redistribution, reproduction** … are prohibited unless written permission has been provided by IEA.

Each item page also carries the watermark:

> Copyright protected by IEA. This item may not be used for **commercial purposes** without express permission from IEA.

**Two readings, and they give opposite answers.**

- If "Commercial" distributes across the whole list — *commercial exploitation, commercial distribution, commercial redistribution* — then IRW's non-commercial research use is inside the grant and the table can ship.
- If "Commercial" attaches only to "exploitation", then **all** distribution and redistribution require written permission, and this table must not ship without it.

The extracting agent took the first reading and flagged it rather than resolving it. I think that reading is right: the per-item watermark restricts only *commercial* use, and the preceding sentence affirmatively grants research use — a blanket redistribution ban would make both redundant. But this is third-party copyright and not the pipeline's call.

## The sibling cycles read differently — a ruling here does not transfer

This matters, because the three TIMSS tables came through the same round and it would be natural to treat them together. **They are not the same.**

- **TIMSS 2003** (`cdm_timss03`): "*Although the items are in the public domain*, please print an acknowledgement of the source."
- **TIMSS 2011** (`cdm_timss11`): front matter also states the released items are in the public domain.
- **TIMSS 2007** (`cdm_timss07`): the ambiguous clause above, with no public-domain statement.

So 2003 and 2011 look clean on their own terms and are **not** covered by this hold. Only 2007 is.

## A second, broader question these tables raise

IRW records `License: GPL-3.0` for all three `cdm_timss*` tables. That is the **CDM R package's** licence, covering the response data as the package redistributes it. It says nothing about IEA's item *wording*, which is separate copyright under narrower terms.

[...]

PR #1997, opened 2026-09-06
batch_201: 7 tables shipped, 5 not (HEXACO + gilbert_meta_32 excluded on rights) (#1945)

xingyi-zhang:
Closes out batch_201 under #1945: **8 tables shipped, 4 blocked, all 12 accounted for.** 120 items, 621 rows, 487 labelled option cells.

Batch_201 is the first hand-built batch in the 200+ range, worked backwards from the end of the queue so it stays clear of the round runner (see `itemtext/extraction_batches/round_prompt_v1.md` and #1953).

## What ships

| table | items | rows | verification | basis |
|---|---|---|---|---|
| `oxfordcovid_xue_2024_at` | 44 | 228 | NOT_NEEDED | `data_labels` |
| `paampsmartsud_saba_2023_ders` | 37 (1 blank) | 185 | PARTIAL | `reconstructed` |
| `pezzuti_2025_coolpeople_main_nfc` | 5 | 35 | NOT_NEEDED | `data_labels` |
| `shu_2025_translation_add` | 3 | 14 | PARTIAL | `paper_order` |
| `singh_2025_identity_pba` | 11 | 55 | NOT_NEEDED | `data_labels` |
| `socialstereotype_hughes_2025_judgement` | 12 (6 blank) | 66 | PARTIAL | `reconstructed` |
| `sun_2025_morality_study1_fairnessHEXACO` | 4 | 20 | VERIFIED | `paper_explicit` |
| `sun_2025_morality_study1_fairnessMCQ` | 4 | 18 | VERIFIED | `paper_explicit` |

Seven items ship with deliberately blank `item_text` — `DERS_37_POST`, which is not a scored DERS item, and `affil1`–`affil6`, whose per-code wording is not recoverable. Both are documented rather than guessed.

## `queue_state.csv` — 13 rows

All 12 batch_201 tables move off `pending`: 8 `done`, 4 `blocked`, each stamped with `batch_201` and a timestamp. This is the batch_022 lesson from #1937 applied *before* pushing rather than after — a round takes the first 12 `pending` rows in file order, so leaving them `pending` would have queued them for re-extraction at ~$60 a round.

The 13th row is **`gilbert_meta_39` → `excluded`**, which Ben identified in #1945 as the only genuine cross-register contradiction (`excluded` in both `pending_index_notes.csv` and the sheet sweep, `pending` at `queue_state.csv`). Set directly from the existing note with no extraction attempted, as instructed. It was scheduled for batch_203; folded in here because this commit already touches the file. Say the word and I'll split it out.

## Rights, under the corrected test from #1945

Block only on a quotable licence fee or an explicit no-redistribution clause; silence is permission.

- **DERS** — no fee, and APA PsycTests is behind SSO so nothing is quotable either way. Unblocked; it had been held partly on the superseded stricter reading.
- **HEXACO** — re-checked because `sun_2025_morality_study1_fairnessHEXACO` had already shipped. hexaco.org grants the forms "free of charge, but only for the purpose of non-profit academic research" and restricts only how online administrations are *hosted*. No fee, no redistribution bar. **Worth a second opinion:** the nearest thing to a restriction across the nine published instruments is HEXACO's "must either be password-protected or not searchable through search engines", which governs data collection rather than publishing item text. Flagged, not decided.
- **CVSCALE**, **Behavioral Avoidance Scale** — nothing quotable.

## The DERS numbering trap

Gratz & Roemer's Table III numbers items in the **original 41-item pool**, not the published 36-item numbering the deposit uses; shipping those numbers as-is would have misassigned most of the scale. The published numbering drops pool items 2, 11, 13, 18, 36 and renumbers ascending — and that rule places the article's 11 reverse-scored items on exactly the positions the deposit's own `DERS_SUM_POST` independently pins (90/90 complete rows; best rival single-column drop 42/90; no-reverse scoring 6/90). Subscale coherence is a third route, p = 0.0005.

PARTIAL, not VERIFIED: two items sharing a subscale and a polarity remain interchangeable on the evidence.

## Three corrections to earlier work in this same batch

1. **Both `sun_2025_morality_study1_*` tables were built from the wrong file.** The first pass used `normdat-labels.csv`, whose ids carry an `mv.` prefix, reasoning that `mv.X.Fn` and `it.X.Fn` were the same item "differing only in the rater-frame prefix". They are not the same rendering — `mv.` is the **trait-norming study's** stem-less fragments (`is a fair person`) where Study 1 administered `[target's name] is a fair person.` `Codebook.xlsx` sheet "Study 1" was the right source and also carries a **per-item response scale**, which the old notes had asserted did not exist. `option_text` went from 0 labelled cells to all 38. The two blocks differ at the midpoint (`Neither agree nor disagree` for MCQ, `Neutral (neither agree nor disagree)` for HEXACO); preserved, not harmonised. PARTIAL → VERIFIED.

2. **`pezzuti_2025_coolpeople_main_nfc` is not Need for Cognition** — that was a guess from the code name made while the table was blocked. The `.sav` labels show five items about rules, instructions and standardized procedures: CVSCALE's uncertainty-avoidance dimension. Only 844 of 3,972 respondents saw the block.

3. **`socialstereotype_hughes_2025_judgement`: the codebook settles 6 of 12, not 2.** The first pass read the codebook's variable table and never mined its prose Measures section. `ce1` + reversed `ce2` uniquely reproduce credibility-of-person M = 4.59, SD = 0.71 (the reverse orientation gives 1.41); `nps` separates the business items by sign (+0.447 / −0.572). The scored data's column order (`ce1, ce2, sympathy, ce4, ce5`) also shows `sympathy` is the missing `ce3`. **And `affil6` is not part of the five-item affiliation scale** — the deposit's own scored `affil` composite is reproduced exactly by dropping it, 1211/1211 rows against 221 for all six, so averaging six will not reproduce the paper.

## Two tables unblocked by files supplied directly

`pezzuti_2025_coolpeople_main_nfc` (dead OSF `view_only` token, paywalled article) and `singh_2025_identity_pba` (box URL with a peer-review passcode) were both access-blocked. The deposit `.sav` and ResearchBox #2404 respectively were supplied by hand, turning both into level-1 extractions.

**The dictionary URLs are deliberately left alone**, per @xingyi-zhang: they open in a browser and editing them would affect other documents. An earlier note recommending a fix is withdrawn and now records the decision, so nobody "fixes" it later. For both tables the provenance row points at the reachable source instead.

## Still blocked (4), all determinate

- `wang_onlineriskexp_2025_interpersonalsecurity`, `wang_onlineriskexp_2025_negativeatt` — the deposit and paper publish no item wording in any language (zero CJK across the supplement, so the administered Chinese is absent too).
- `sun_2025_morality_study2_colleaguerespect`, `sun_2025_morality_study2_colleaguetrust` — the `_1..7` suffix indexes **informants, not items**, stated in the deposit codebook and confirmed by non-missing counts correlating +0.985 with `cov_teamsize`. This is a response-table shape question, not an item-text gap; happy to file it separately if it should be pursued.

Confirmed as genuinely blocked by @xingyi-zhang. These do not count toward the circuit breaker.

## Filed separately, not fixed here

**#1996** — `data/oxfordcovid_xue_2024.r:204` filters the `_at` table on `resp == floor(resp) & resp <= 5`. Its eight `activity_*`/`media_*` items are free-text *hours per day* and the prompt invites decimals, so the filter deletes 12,985 responses (23% of those eight items), 89.6% of them merely for being non-integer; the raw median is 0.5, i.e. the modal answer type. Item text is left as-is by decision — it gets re-uploaded when the response table is fixed.

## Gates

- `validate_items.R` — PASS on both axes for all 8
- `audit_batch.R` — 5 PASS, 3 WARN; all three WARNs are disclosed deliberate blanks (`oxford`'s 8 quantity items, `DERS_37_POST`, `hughes`'s 6 affil items)
- `normalize_nulls.R` — clean on 8
- `lint_verification.R` — clean on 8. It correctly caught an attempt to use `NOT_NEEDED` with `paper_explicit`; a codebook workbook is not the data file's own labels, so those two rows are VERIFIED instead.
- `check_provenance.R` — 0 entries
- `irw-validate` — exit 0, with the 2 pre-existing `name_charset` WARNs on the capitalised `sun_*` warehouse table names

**One caveat on how the live gates ran.** `validate_items.R --table-sets` hung for 15 minutes on a Redivis server-side call (2.9s CPU, two idle ESTABLISHED connections), and it stayed hung on later attempts. `validate_items.R` and `audit_batch.R` were therefore run against this session's own cached `irw_fetch()` results via `--resp-csv` / `--resp-dir` — same data, no export spend, but worth re-running live if that matters to review. The two `verify_sun_*` scripts fetch live and hang for the same reason; their numbers were reproduced from the cache.

Ref #1945. Uploads are Ben's, as always — nothing here has been uploaded.


ben-domingue:
Ben ruled on both escalations in this batch: **"let's err on the side of *not* having things. so do not host."**

Pushed as b8d240c on this branch.

- **`sun_2025_morality_study1_fairnessHEXACO` removed from batch_201** — the two files and its rows in `provenance.csv`, `verification_merged.csv` and `audit_report.csv`. Nothing had been uploaded (`uploaded` was empty), so this is a pre-merge removal, not a withdrawal. The batch is now 7 tables and all four artifacts agree at 7. Title updated to match.
- **`gilbert_meta_32`** — `pending` → `excluded`, never extracted.
- **`de_vries_2022_hexaco_{self,other,meta}`** — `blocked` → `excluded`. Their batch_026 notes named a ruling from Ben on this exact hexaco.org clause as the one thing that would change them; it has now been given, so they stop being an open gap that future rounds retry.
- **`face_memory_test`** stays `blocked` — the ruling settles the HEXACO half, but the image/asset block on items 1–75 is independent and unresolved.

Reasons recorded in `pending_index_notes.csv` for each row, so the next round reads the decision rather than re-deriving it.

**One thing this does not do.** `sv-maia2_randelovic_2021_hexaco60` and `_hexaco100` are **live** with published HEXACO-PI-R wording, uploaded before #1891 existed — the class escalation already logged in `round_log.md` on 2026-09-04. This ruling is the answer that escalation asked for, but taking down published item text is a withdrawal and needs Ben's explicit go-ahead, so it is left alone here. `dasilva_2019_hexaco24` is unaffected (Brief HEXACO Inventory, CC BY).

Otherwise ready to merge.

https://claude.ai/code/session_014r1psZ76WGKjmiQUNLaFdT

ben-domingue:
Pushed `0009303` to this branch — `oxfordcovid_xue_2024_at`'s item text had to shrink, because the response table is losing eight items.

## Why

#1996: `data/oxfordcovid_xue_2024.r:204` filtered the `_at` table with `resp == floor(resp) & resp <= 5`, which deleted 12,985 responses — all of them from the eight `text` items you documented as average hours per day. Since those items invite decimals, the integer test deleted 23% of exactly the answers the question asked for. Ben's call was to **drop the eight rather than repair them**: they're unrelated time-use quantities, not a scale, and not the kind of data IRW carries. Script fix is #2008.

Your read of the table was right in every particular — the two response formats, the `Field_Type` split, the deliberately blank `option_text`, and the note that the `resp` 0–5 range was an artifact of line 204 rather than a response scale. That diagnosis is what made the fix straightforward.

## What changed here

`oxfordcovid_xue_2024_at__items.csv`: **228 rows / 44 items → 180 rows / 36 items.**

The two sections split cleanly along the two formats — section `_1` was exactly the eight hours items, `_2` exactly the 36 reason items — so this is a whole-section deletion. Every retained line is byte-identical apart from `_2` → `_1`, renumbered to match every other single-section file in the batch. Consequences:

- `option_text` populated on **all** 180 rows, so the blank-`option_text` half of `audit_batch.R`'s WARN is resolved.
- `resp` collapses to 1–5; `resp = 0` occurred only on the hours items.
- `public_note` no longer needs to explain two formats. It now carries the thing a user actually needs: each medium's six reasons were shown only to respondents who used that medium, so item n varies by medium.

`provenance.csv` (note, public_note) and `verification_merged.csv` (evidence) rewritten for the 36-item table. `audit_report.csv` **left as-is on purpose** — it's the generated record of the audit that ran against the 44-item live table, and re-running it cleanly isn't possible until #2008 merges and the table is re-uploaded.

## One ordering constraint

Item and resp sets must match live data. So this table's item text goes up **only after** #2008 is merged and `oxfordcovid_xue_2024_at` is re-uploaded at 36 items / 136,467 rows. Uploading before that would ship an item table disagreeing with the live one.

Nothing here is uploaded yet — `uploaded` is `NA` for all seven tables in both sidecars — so there's nothing to correct, just an order to respect. The other six tables in this batch are unaffected and can go whenever.

ben-domingue:
@xingyi-zhang let me know if i need to upload stuff? just email me a zip file?

PR #1998, opened 2026-09-06
Rule HEXACO-PI-R out of the corpus; stage the two live withdrawals (#1945)

ben-domingue:
Ben's #1945 ruling — *"let's err on the side of *not* having things. so do not host."* — as a durable rule, plus the answer to the HEXACO class escalation the round log raised on 2026-09-04.

The pre-merge half is on the batch_201 branch (b8d240c, #1997). This PR is the rule and the staged takedown.

**Why the rule cannot be re-derived.** batch_201 shipped a HEXACO table after testing hexaco.org's *"free of charge, but only for the purpose of non-profit academic research"* against the 2026-09-04 rulings and finding it carried neither a quotable fee nor a no-redistribution bar — the two things those rulings made decisive. That test was applied correctly; the ruling is that passing it is not enough. `itemtext_standard.md` records the superseded reading explicitly, because a future round applying the fee/redistribution tests alone will get this wrong.

**Not everything with HEXACO in the name is in scope.** `dasilva_2019_hexaco24` is the Brief HEXACO Inventory (CC BY), untouched. The other 26 live `sun_2025_morality_*` item-text tables are the study's own moral-character template — traced to `itemtext/itemtables/pilot/audit_confirmed.csv` — which is why the one HEXACO block in that deposit is the table the deposit itself named `fairnessHEXACO`. Response tables are out of scope entirely: the clause governs instrument wording, not data collected with it. That is the difference from `wvs_panasiuk`, where the microdata was the problem; the precedent here is PROMIS.

## The takedown has run

`sv-maia2_randelovic_2021_hexaco60` and `_hexaco100` were deleted from the `irw_text` draft by Ben on 2026-09-06 (`tools/withdraw_hexaco.py`, `APPLY=1`). Verified: the draft went 732 -> 730, neither target is present, and `dasilva_2019_hexaco24__items` -- the Brief HEXACO Inventory, CC BY, the one HEXACO-named table that stays -- is still there. Their rows are out of `metadata/itemtext_metadata.csv`.

The wording is still in released **v19.0** and leaves only when the draft is released, so until that cut the tracked baseline describes the draft rather than the corpus. Recoverable until then, not after.

No `itemtext_issues.qmd` entry, per the 2026-09-05 rule that withdrawals are not published.

Merged with main on 2026-09-08: `itemtext_standard.md` (this branch's HEXACO section vs main's `wording_rights` retirement) and `pending_index_notes.csv` (2 rows vs 39) both conflicted only on the append anchor; both sides kept. `metadata/itemtext_metadata.csv` auto-merged onto main's regenerated baseline with the two deletions intact. Both CI jobs green.

Refs #1945.

https://claude.ai/code/session_014r1psZ76WGKjmiQUNLaFdT


ben-domingue:
Correction to my own scoping claim, pushed as 55017ac.

I wrote that the other 26 live `sun_2025_morality_*` item-text tables are the study's own moral-character template rather than HEXACO-PI-R, and stated it as settled. It is not settled — it rests on the deposit naming its one HEXACO block `fairnessHEXACO` plus the note in `itemtext/itemtables/pilot/audit_confirmed.csv`. The live fetch that would have read the actual wording failed on the Redivis outage and has not been re-run.

The specific residual risk is Study 3, which ships `extraversion`, `openness` and `neuroticism` — personality factor names, not moral-character ones. Weak evidence against HEXACO: `neuroticism` is a Big Five label, where HEXACO's factor is `Emotionality`.

Now recorded in the standard as an open check to run alongside the withdrawal, rather than as a conclusion. If any of those 26 do hold HEXACO-PI-R wording, they fall under the same ruling.

https://claude.ai/code/session_014r1psZ76WGKjmiQUNLaFdT

Issue #1863, opened 2026-09-03
03_tags.R supersedes at row level where column level was wanted (19-76 tables per column unreachable)


ben-domingue:
Child of #1704 (item 2). Found while sizing the untagged population on
2026-09-03.

### The bug

`03_tags.R` implements #1723's human-wins rule at **row** level: when a table has
a row from the tags Sheet, the entire automated row for that table is dropped —
not merely the columns the human filled.

So a table whose Sheet row carries `NA` in `construct type` can never receive an
automated `construct type`, because the automated row never survives long enough
to be merged. The human "won" a cell they left empty.

### Scale

Measured 2026-09-03 against `metadata/metadata.csv` and `metadata/tags.csv`,
joined case-insensitively, treating the literal string `NA` as blank:

| column | Sheet-held with `NA` — permanently unreachable |
|---|---|
| `primary language(s)` | 76 |
| `construct type` | 28 |
| `measurement tool` | 21 |
| `item format` | 19 |

Small next to the 1,427 reachable untagged tables, and deliberately not a
blocker on the coverage push. But it is a permanent ceiling — no tagging run of
any size can ever fill those cells — and it grows as the Sheet grows.

### Why this is worth recording carefully

It has been repeatedly misattributed, including in this project's own docs, as a
**Google Sheets write-path problem** — i.e. something requiring a service account
and blocked on #1708. It is not. `ARCHITECTURE.md` §3 settles that no service
account will be provisioned, and #1723 already routes automated tags around the
Sheet entirely via `tags/tags_auto.csv`.

This is a merge-granularity bug in one R script. No Sheets API, no credential, no
new dependency.

Sub-action 2.4 in #1704 ("Fix the write path. **Depends on 6.1 (#1708)**") is
stale for the same reason — it was written before #1723 shipped.

### The fix

Merge column-wise rather than row-wise: a human value wins the cell it occupies;
an automated value fills a cell the human left blank. Guards worth keeping:

- Only fill blanks. Never overwrite a non-empty human cell.
- `NA` in the Sheet must be treated as blank, not as a value — that is how it is
  stored today.
- The derived-age precedence hole (#1760 decision 7) is unaffected: it already
  outranks both Sheet and auto rows for `age range` and `child age` only.
- `metadata/tags_provenance.csv` should record cells filled this way, as it
  already does for other automated values, so they remain sweepable and
  revertible.

### Carry the same lesson to #1732

6b proposes giving the dictionary the same union treatment. **It should be built
[...]

PR #2000, opened 2026-09-06
dictionary: give it the same write path as tags, column-wise (#1732)

### LINKED ISSUE #1732: [6.4] Give the dictionary the same write path as tags
Sub-action **6.4** of #1708. **Queued behind 2.4 (#1723) — do not start before it
ships and has run a cycle.**

### The dictionary has an automated writer, and nobody designed its write path

6.1 was resolved for tags: automated rows go to a git-tracked CSV, `03_tags.R`
unions it with the sheet export, humans win on conflict. The dictionary was left
on business-as-usual — on the implicit assumption that only humans write it.

That assumption is false. `automated_finding/` currently holds **five** scripts
that generate dictionary rows, all written between 2026-08-27 and 08-29:

    make_biblio_2026-08-27.py
    make_biblio_tierB_2026-08-27.py
    make_biblio_zenodo_2026-08-28.py
    make_biblio_zenodo_backlog_2026-08-29.py
    make_biblio_2026-08-29b.py

Each hardcodes one batch's rows and emits a CSV whose docstring reads *"Fully-
quoted, LF-terminated, ASCII .csv for File > Import > Append (comma)"* — output
shaped for a human to paste. `NAMING_AUDIT_HANDOFF.md` records the sheet growing
by ~200 automated rows in a single day, and `TODO.md` logs the paste step
repeatedly ("uploaded to Redivis and pasted into the dictionary sheet").

Humans also add dictionary rows. **Two writers with no precedence rule** is
exactly the problem the tags union solves.

### What this buys

- A reviewable diff for automated rows. ~200 rows currently land with no diff, no
  blame, and no revert.
- One pattern across both sheets instead of two, and a schema check at write time.
- An end to writing a new bespoke script per batch.

### Blocker to resolve first: the dictionary schema has duplicate column names

`Custom License` appears **twice** in the dictionary layout, at positions 8 and
11. `metadata/hotfixes/fix-licenses.R` reads a column named `Custom License...11`
— R's `make.unique` silently disambiguating. A name-keyed union is ambiguous
against that schema.

Decide before building: fix the duplicate in the sheet, or select positionally as
`03_tags.R` does with `KEEP_COLS`. Positional selection is already the most
fragile thing in that file, so prefer fixing the schema.

### Why it waits for 2.4

- **Blast radius.** `tags.csv` feeds one Redivis table. The dictionary feeds
  `biblio` for all four sources, is read at nine-plus call sites, and is
  `automated_finding`'s live dedup-exclusion source — a bad export there changes
  which datasets get discovered.
- **2.4 is the prototype.** Every precedence edge case surfaces there first and
  cheaply. Generalising before the first instance runs produces an abstraction
  that fits neither case.

### Possible connection to #1690

Those scripts write a Zenodo **dataset** DOI into the `DOI (for paper)` column —
exactly the defect #1690 describes across 465 rows. Stated at its real strength:
they cover ~120 tables and #1690 predates them, so they are a demonstrated
instance of the pattern, not proven to be its whole cause. Worth checking when
6.4 is picked up, since a validated pipeline input would catch this class of
error in a diff rather than 465 rows later.

### Done when
Automated dictionary rows reach `biblio.csv` through a reviewed, git-tracked
input with human rows winning on conflict, and the per-batch `make_biblio_*.py`
scripts are retired.

Found while writing ARCHITECTURE.md (#1726), which had to describe two sheets
with two different write paths and could not say why.

ben-domingue:
Closes #1732 (roadmap item 6b).

Automated dictionary rows reached the sheet by hand — a per-batch
`make_biblio_*.py` emitted a CSV and a human ran File > Import > Append. ~200
rows landed that way in one day with no diff, no blame and no revert. Humans
write the same sheet directly, so there were two writers and no precedence rule.

Automated rows now go to `automated_finding/dictionary_auto.csv` (git-tracked,
written by `stage_dict_row.py`), and `metadata/02_biblio.R` unions it into the
sheet export on every run.

### Column-wise, not row-wise

`03_tags.R` drops the whole automated row when a human row exists, so a human
who left a cell blank still wins it — 19–76 tables per column stranded (#1863).
For the dictionary a sparse-but-present row is the *common* case, so row-level
supersede would mean an automated writer could only ever reach a table with no
dictionary row at all.

Here a human cell wins the cell it occupies; an automated cell fills a cell the
human left blank, with `"NA"` counting as blank because that is how the sheet
stores it.

### Two things the tags prototype got wrong, and this does not

- **`metadata/biblio_provenance.csv` is committed.** `tags_provenance.csv` is
  excluded by `.gitignore`, so the weekly runner regenerates and discards it —
  no record survives of which cells a machine wrote.
- **`tests/test_dict_union.R` runs in CI** (28 assertions). The load-bearing one
  is that a human's non-empty cell is never overwritten.

### A correction worth reading

**The two `Custom License` columns are not both blank.** Position 8 holds 20
non-blank values (the source's own terms — Stanford ROAR, Harvard Dataverse
custom terms) and position 11 holds 4 (the terms IRW redistributes under —
GroupLens' full text), all on Public tables, verified 2026-09-06.

`SKILL.md` said "(both blank)" and #1732 proposed **deleting one**. That would
have destroyed licence text. `resolve_dict_cols()` disambiguates them by order
today and by name once the sheet is renamed; neither is deleted, and the
`SKILL.md` claim is corrected.

### Custom licence terms now reach users

238 Public tables publish `Derived_License = "Custom"` — the bare word, with no
terms anywhere a reader can reach. `hotfixes/fix-licenses.R` added the column
once (Rpkg#93) and the next pipeline run erased it, because `02_biblio.R`
rebuilds `biblio.csv` from a fixed column list.

`Custom_License_Terms` is now carried for **every** row, not just new ones — the
tables holding terms are long-published, so a new-rows-only carry-through would
deliver them to nobody. 4 rows gain terms; 3 of the 238. The other 235 have no
terms recorded anywhere, which is a data gap, not a code one.

### Retired

Six `make_biblio_*.py`, the File > Import > Append procedure in `SKILL.md`, and
`hotfixes/fix-licenses.R`.

### Verification

- `Rscript tests/test_dict_union.R` from `metadata/` — 28 assertions, all pass,
  offline.
- Round-tripped a row through `stage_dict_row.py` → `read_dict_auto()` with a
  comma, a quote and a newline in `Description`.
- Licence-terms join run against the live sheet and the current `biblio.csv`:
  4 rows gain terms, no other column moves.
- Nothing here uploads. `upload_meta.py biblio` and the Publish click stay
  separate and manual.

### Not addressed

`02_biblio.R` still never re-reads an existing row's other columns, so **226
tables publish a Description, DOI or Reference the dictionary contradicts**.
Filed separately — it moves published rows and deserves its own decision.

https://claude.ai/code/session_01CFNUSQXJ4jfrdAbKmHebo6


ben-domingue:
## Test plan

Three tiers. A and B are scripted and run against production data without
touching anything; C is the real batch and is the only one that costs money.

### Tier A — the merge rule against the live sheet

```bash
cd metadata
Rscript tests/manual_dict_test.R tier-a
```

Reads the live dictionary (4,384 rows), picks a real live+Public table whose row
has a blank DOI, stages a row through `stage_dict_row.py` for real, and asserts
the human's `Description` survives while the blank DOI fills. Currently passes on
`16_personalityfactors`.

Catches what fixtures cannot: the duplicate `Custom License` names, and the real
mix of `NA` / `""` / `"NA"` that raters type.

### Tier B — the export path replayed on the real biblio.csv

```bash
Rscript tests/manual_dict_test.R tier-b
```

Offline replay of what `02_biblio.R` does to existing rows. The load-bearing
assertion is that **all 7 pre-existing columns are byte-identical** — the change
is additive. Also reports the licence-terms result (4 rows gain terms; 3 of the
238 `Custom` tables) and demonstrates the held-row window.

Neither tier writes outside `tempdir()`, and neither touches Redivis, Anthropic,
the sheet, or the tracked `dictionary_auto.csv`.

### Tier C — one real batch

Deliberately small: two or three `good` candidates, not a full sweep. The point
is to exercise the path, not to grow the corpus.

1. `irw_discover_updated.py` with a narrow term → `irw_batch_updated.py --limit 10`
2. Process 2–3 `good` rows into IRW tables as normal
3. Upload the tables to a Redivis draft
4. **Stage the dictionary rows with `stage_dict_row.py`** — this is the step
   under test, replacing a `make_biblio_*.py`
5. Run stage 02 **before** publishing. The rows should be *held*, and
   `metadata/biblio_pending.csv` should name them
6. Publish the tables, re-run 02. The rows should land, and
   `biblio_provenance.csv` should name the cells they supplied
7. Upload `biblio` and publish

Watch for: refusals firing at staging time rather than at export; a
`Description` containing a comma or quote surviving the round trip (the old
paste path's failure mode); `comps`/`nominal`/`simsyn` biblio outputs unchanged;
and BibTeX generation on the new rows, which needs `ANTHROPIC_API_KEY` and is
the one step that costs money.

### Afterwards

No new tracker needed. `metadata/biblio_provenance.csv` is committed, so
`git log -p metadata/biblio_provenance.csv` is the permanent record of every cell
the machine supplied and when — which is exactly what the tags side cannot tell
you, because its sidecar is gitignored.

The question worth answering over the next few weekly cycles: **does a human ever
overwrite a cell the machine filled?** If yes, the machine's values are landing
wrong. If never, the column-wise rule is doing nothing a simpler append wouldn't,
and the insurance isn't needed. Both answers are useful, and both come from
diffing that one file week over week.

https://claude.ai/code/session_01CFNUSQXJ4jfrdAbKmHebo6

Issue #1942, opened 2026-09-04
ENEM ingest: blanks scored as wrong answers, enem_2019_1mil_lc looks mis-scored, and no post-scoring sanity check


ben-domingue:
Three related findings in the ENEM ingest (`src/data/enem_<year>.R`), surfaced while building the guessing vignette in `irw_site`. The first is a definite one-line bug affecting every ENEM table in IRW. The second is a specific table that looks mis-scored. The third is the guard that would have caught both at build time.

## 1. Blanks and double marks are scored as wrong answers

All 13 year scripts (`enem_2013.R` … `enem_2025.R`) share this pair of lines, at line 9 and line 121 of each:

```r
# INEP response codes: A-E, "." blank, "*" double mark
ALPHABET <- c("A", "B", "C", "D", "E", ".", "*")
...
out <- df |> mutate(resp = if_else(raw == key, 1, 0))
```

The script knows `.` and `*` are valid non-responses — it validates the raw codes against that alphabet and stops the build on anything outside it — and then scores both as incorrect. A candidate who did not sit a section has a full block of blanks, and every one becomes a `0`.

**Observed consequence.** In a fixed-seed 3,000-candidate draw from each table, the share of candidates scoring *exactly zero with every item non-missing* is:

| table | share at zero | chance floor |
|---|---|---|
| enem_2013_1mil_mt | 45.3% | ~9 of 45 |
| enem_2013_1mil_lc | 44.7% | ~10 of 50 |
| enem_2013_1mil_ch | 44.6% | ~9 of 45 |
| enem_2013_1mil_cn | 43.9% | ~9 of 45 |
| enem_2014_1mil_ch | 44.6% | ~9 of 45 |
| enem_2019_1mil_ch | 0.03% | |
| enem_2024_1mil_ch | 0.13% | |

Scoring 0 on 45 answered five-option items has probability ~4e-5, so these are not low-ability candidates. The split is an absence-rate difference between administrations, not a data-quality gradient across tables.

**Downstream effect.** Left in, the block inflates the estimated latent SD to 2.2–2.9 and drives every item's guessing parameter to the no-guessing boundary, because a floor model cannot fit observed zeros where it predicts a floor. Screened out, SD(θ) falls to 0.4–0.7 and the floors become estimable. In the guessing vignette this reversed the headline finding: what looked like "these tables split in two, and guessing corrections hurt on half of them" was one coding artifact seen twice.

**Proposed fix**, in each year script:

```r
[...]

PR #2020, opened 2026-09-07
ENEM: stop scoring blanks and double marks as wrong answers (#1942)

ben-domingue:
Fixes findings **1** and **3** of #1942. Finding 2 (`enem_2019_1mil_lc`) is not fixed here — see below.

**Nothing is rebuilt.** The INEP microdata is not on this machine, and the blast radius is large: **52 live tables, 2.34 billion rows**. So this is the code change, verified as far as it can be without the source files, and the rebuild is its own operation.

## Finding 1 — the one-line bug

```r
resp = case_when(raw %in% NONRESPONSE ~ NA_real_,   # "." blank, "*" double mark
                 raw == key           ~ 1,
                 TRUE                 ~ 0)
```

Dropped **after** the raw-code validation, which still needs to see them, and with a count printed per area.

Worth stating because it changes who can work around this: **`resp_raw` is not published.** The live tables carry only `id / item / resp / position / booklet`, so a consumer cannot screen the non-responses out themselves. Everyone pulling these tables gets the 45%.

`*` is a judgement call and is written down as one, per the issue: a double mark is a response that could not be read, which is not evidence of a wrong answer either. `NONRESPONSE` is the single line to change if a later analysis wants them scored.

## Finding 3 — the guard, and it is tested

`data/enem_checks.R`. Every existing guard in these scripts is on the *inputs and the joins*; both defects produce structurally valid output that is psychometrically impossible, so none of them fires. The new checks run on the **scored output**:

| condition | action |
|---|---|
| an item scored identically for every candidate | **stop** — `enem_2019_1mil_lc` has five, 0 correct out of 7,520 |
| median item-rest < .05 *and* >60% of items below it | **stop** — a correctly keyed item correlates positively with the rest regardless of difficulty |
| >50% of items below the 1/5 chance floor | warn |
| >5% of candidates who answered every item scoring exactly zero | warn — this is finding 1 |

Item-rest on 45M rows is not free, so it runs on a seeded 5,000-candidate sample; p-values use everything.

`data/tests/test_enem_checks.R` exercises it offline on synthetic forms built to look like each defect, and — the test that matters — on a **legitimately hard form**, which passes. Difficulty alone must never fail a build. Added to `.github/workflows/tests.yml` (base R only, no microdata, no network), so the thresholds that decide whether a 45-million-row table is publishable are checked on every PR rather than only when someone has 13 years of INEP files on disk.

```
a healthy form passes                        4 ok
finding 1 -- non-responses scored as wrong   3 ok
finding 2 -- a mis-keyed form                5 ok
shape                                        2 ok
all tests passed
```

## What I could and could not verify

- **Could:** the 13 scripts still differ only in their two prova-code lines, exactly as before — the same patch applied 13 times, mechanically, with no drift. All 14 files parse. The guard is tested against both defect shapes and against a false positive.
- **Could not:** run any year script. That needs the INEP microdata. So the scoring change is verified by inspection and by the identical-diff property, not by execution against real data.

## Not in this PR

**Finding 2, `enem_2019_1mil_lc`.** Still needs `ITENS_PROVA_2019.csv` and the 2019 microdata to confirm whether the LC language items sit where the `width == npos + 5` branch assumes. The new guard would have *failed* that build — five zero-variance items stop it outright — but diagnosing it is a separate job. Note that this table stays wrong-now until then; the finding-1 fix does not touch it.

**The docs line.** There is no ENEM markdown to put it in; the reasoning now lives in the script header, which is where a reader of these scripts would look. The user-facing statement belongs in the dictionary `Description` for the 52 tables, which is a sheet edit rather than a code change.

## Note on the refactor

The issue suggests factoring the 13 near-identical scripts into one helper. I did not, deliberately: I cannot execute them, and collapsing 13 files I cannot run is a bigger change than the bug warrants. The patch being provably identical across all 13 is the next best thing. Happy to do the refactor as its own PR once there is a machine that can run one year end to end.

https://claude.ai/code/session_01UjpFgj7bT9GJ8UeF5hMLdE

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -83,6 +83,14 @@ jobs:
         working-directory: metadata
         run: Rscript tests/test_dict_union.R
 
+      # The ENEM post-scoring checks (#1942). Base R only -- no microdata, no
+      # network -- so the thresholds that decide whether a 45-million-row table
+      # is publishable are exercised on every PR rather than only when someone
+      # has 13 years of INEP files on disk.
+      - name: enem scoring checks
+        working-directory: data
+        run: Rscript tests/test_enem_checks.R
+
   # A PR to data/ contains a conversion SCRIPT, never a table --
   # automated_finding/irw_output/ is gitignored, so no output is ever in the
   # diff. This checks the script's contract; the data is checked by
--- a/data/enem_2013.R
+++ b/data/enem_2013.R
@@ -8,6 +8,25 @@ regular_prova_codes  <- c(167, 170, 171, 174, 176, 177, 178, 180, 181, 182, 189,
 # INEP response codes: A-E, "." blank, "*" double mark
 ALPHABET <- c("A", "B", "C", "D", "E", ".", "*")
 
+# "." and "*" are NON-RESPONSES, not wrong answers (#1942).
+#
+# This script validated the raw codes against ALPHABET and then scored both of
+# them 0, so a candidate who did not sit a section came out with a full block of
+# zeros. In a 3,000-candidate draw from the 2013 and 2014 tables, 44-45% of
+# candidates scored exactly zero with every item present -- on 45 answered
+# five-option items that has probability ~4e-5 each, so the block is coding, not
+# ability. Left in, it inflated SD(theta) to 2.2-2.9 and pinned every item's
+# guessing parameter to the no-guessing boundary.
+#
+# `.` is unambiguously a blank. `*` is a double mark -- a response that could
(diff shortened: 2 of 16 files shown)


ben-domingue:
Closing unmerged. #1942 is @mateusmazza's call, and this got ahead of that — the ENEM ingest is his, and the scoring change moves 52 published tables.

Nothing here is lost: the findings are written up as notes on the issue. The branch `fix/1942-enem-blank-scoring` stays pushed if any of it is useful as a starting point, but it should not be merged on my account.

Issue #1955, opened 2026-09-05
wording_rights: NC cannot express a fee, and the flag is now set on zero tables


ben-domingue:
`itemtext_standard.md` § Rights says that where a rights holder states an instrument-level restriction, every row of that table gets `wording_rights=NC` plus a public issues-page entry. The column is meant to be a filterable flag so a commercial reuser can exclude those tables with a query.

Two problems surfaced on 2026-09-04.

**1. `NC` is the wrong label for a fee.** The TAS-20 ruling blocks instruments whose rights holders levy an **enforced licence fee**. A fee is not a non-commercial clause — it is a price, and it applies to commercial and non-commercial reuse alike. Marking such a table `NC` would tell a reuser something false. The three tables withdrawn on the fee rule were therefore left with the column **unset**, which is honest but leaves no flag at all.

**2. The flag has never had more than one user.** `CV_OASIS_ODSIS_PPE_Novak_2020_DSES` was the only table in the corpus carrying the column, and it has now been withdrawn — so `wording_rights` is currently set on **zero** tables. Any sweep that tries to use it to find restricted tables (see the re-audit issue) will find nothing and may wrongly conclude there is nothing to find.

**What needs deciding.** Either extend the vocabulary (e.g. `NC`, `fee`, `no_redistribution`) so the flag can express the restriction that actually applies, or drop the column and rely on `public_note` and the issues page alone. Leaving it as a one-value flag that nothing currently carries is the worst of the three.

Whichever way it goes, `itemtext_standard.md` § Rights needs updating — it currently documents only the `NC` value.

Context: #1709.

ben-domingue:
Ruled 2026-09-06: **drop the column**, and withdraw the tables it covered. PRs #2006 (repo) and datapages/irw#145 (site).

### The flag was set on eight tables, not zero

Measured against `irw_text` v18.0 by listing variables for all 760 live item-text tables. `CV_OASIS_ODSIS_PPE_Novak_2020_DSES` was the *first* table to carry `wording_rights`, not the only one — and six of the eight below were uploaded after the count that produced "zero", on the evening of 2026-09-04 and on 09-05.

| table | instrument | the rights holder's clause |
|---|---|---|
| `alsuhibani_2022_ecrs_s3` | ECR-12 | "You may not use the scales for commercial purposes without permission" |
| `conspiracy_asd__asd_aq10` | AQ-10 | "used for research purposes and not for commercial use" (ARC Cambridge) |
[...]

Issue #1927, opened 2026-09-04
Decision: WHOQOL-BREF item wording rights, corpus-wide (7 tables, 3 already published)


ben-domingue:
Raised by batch_020 triage (#1709). **This is a decision for @ben-domingue, not a defect.** Nothing is blocked on it today; I am filing it because it should be ruled once rather than re-argued per table, and because three tables are already live.

## The question

WHO and the UK WHOQOL centre operate a **user-agreement scheme** for the WHOQOL instruments, reportedly royalty-free only for non-commercially-funded research. If that is a stated non-commercial restriction, the #1891 ruling would bar shipping the item wording.

## Why it was not applied in batch_020

#1891 fires on a **quotable** restriction only — explicitly "never on an inference that a scale is copyrighted or reproduced without an explicit grant, which was considered and rejected as too broad." No such sentence could be retrieved:

- who.int's WHOQOL-BREF page states no licence terms
- cpcr.aut.ac.nz returned 403
- the Manchester user-information PDF failed TLS

The wording shipped in `COACH_Chen_2022_WHOQOL_BREF` was also taken from a **CC0** Harvard Dataverse deposit, and under the 2026-09-04 source-licence ruling that is what governs. So `wording_rights=NC` was correctly **not** set, and the table is staged.

## Scope, which is the reason to rule now

Seven WHOQOL tables are in the corpus:

| table | itemtext status |
|---|---|
| `altahla_2024_whoqol` | **published** 2026-08-17 |
| `altahla_2024_whoqol_bref` | **published** 2026-08-17 |
| `burkert_2019_whoqol_bref` | **published** 2026-08-18 |
| `COACH_Chen_2022_WHOQOL_BREF` | staged in `clean/`, batch_020 |
| `alkhaldi_2023_whoqol_autism` | not yet extracted |
| `galgam_2025_whoqol_bref` | pending |
| `silva_2018_whoqol` | pending |

So the wording is **already live for three tables**. Holding the fourth would not reduce exposure; it would only make the corpus inconsistent. Whatever is decided applies to all seven.

## What would settle it

A verbatim rights-holder sentence, one way or the other — most likely from the WHOQOL user agreement itself rather than from who.int. If it says non-commercial, #1891 fires and the three published tables need their wording withdrawn. If it does not, the current position stands and this closes.

Refs #1709, #1891.

ben-domingue:
[...]

Issue #1643, opened 2026-08-18
Itemtext: hold decision on ALSECYPIAMH_WU_2022_PHQ (unverifiable 2-item mapping)


ben-domingue:
The itemtext for this table is extracted and validated but **held, not uploaded** (itemtext/itemtables/batch_004/).

**Why it is held.** The mapping rests on the convention that column N is instrument item N. That same convention was accepted for `almuqbil_2022_epds` and `altahla_2024_swls`, which shipped. The difference here is that PHQ-2 has only two items (anhedonia vs depressed mood), so a wrong guess affects the whole table, and there is no marker item to corroborate it.

**What was ruled out** (recorded in itemtext/mapping_verification.csv):
- The OSF raw file `CPS Study 2.sav` labels 39 of its 92 columns, but not PHQ1/PHQ2.
- The OSF supplementary docx reproduces only the Claremont Purpose Scale.
- The source paper (10.1111/jora.12944) is paywalled and absent from Europe PMC OA.
- Live means (PHQ1 0.939, PHQ2 0.816) distinguish the items but give no anchor for which canonical item should be higher.

**Decision needed:** ship with its public note, or drop. Resolvable by author contact or paywalled-paper access.

xingyi-zhang:
from the paper:

Then, the 2-item Patient Health Questionnaire (PHQ-2; Liao et al., [2017]); Löwe et al., [2005]) assessed participants' negative affect. The PHQ-2 measures individuals' presence of depressive symptoms using a 4-point scale ranging from 0= “not at all” to 3 = “nearly every day”

this is form the Liao paper: 病人健康问卷 2 条目中文版（PHQ-2-C） PHQ-2-C 量表是 9 个条目的抑郁自评问卷，病
人健康问卷 9 条目的前 2 个条目的中文版，其主要 判定过去两周内的感受改变情况［4］。问卷中条目 1 为“做事情没什么兴趣”，条目 1 为“感到沮丧、压抑或绝望”，每个条目分为 0～3 级评分，其中 0 分为未出现，1 分为偶尔几天会出现；2 分为两周中一半以上时间出现；3 分为几乎每天都出现.

since the item names are PHQ1 and 2, should follow the same order.

ben-domingue:
Hold released — thanks, that supplies exactly what the earlier `NO_ROUTE` couldn't reach.

[...]

PR #2033, opened 2026-09-07
itemtext: read liveness from Redivis, not from the `uploaded` column (#1828)

### LINKED ISSUE #1828: provenance `uploaded` is a hand-maintained mirror of something Redivis already knows
Found while clearing the last of #1816. **16 provenance rows name a table that is live in `irw_text` and leave `uploaded` blank.**

| file | rows |
|---|---|
| `itemtables/batch_015/provenance.csv` | 11 |
| `itemtables/batch_012/provenance.csv` | 2 |
| `itemtables/batch_004/provenance.csv` | 1 |
| `itemtables/batch_011/provenance.csv` | 1 |
| `itemtables/batch_014/provenance.csv` | 1 |

The 11 in `batch_015` are already accounted for — they went live in v14.0 today and PR #1820 will stamp them. The other five are older and nobody knows when they went up.

## Why I did not just stamp them

The honest date is not recoverable. Redivis' per-table `createdAt` reports **2026-09-02** for all five, which is wrong — opening a draft copies the whole dataset and resets those timestamps. It is the same trap `red_up.drafts` documents for measuring draft age: a table's `updatedAt` and the draft version's `createdAt` both reset whenever the draft is touched.

So the choice was to write a date I know to be wrong, or leave it blank. A provenance record with a fabricated date is worse than one with a gap, so: blank, and this issue.

## The underlying problem

`uploaded` is a **hand-maintained mirror of a fact Redivis already holds**. Whether a table is live is not an opinion; it is answerable in one API call. Every one of these 16 rows is someone having done the upload and not come back to the CSV — which is the failure mode `ARCHITECTURE.md`'s Rule 2 is about: *prefer documentation that cannot go stale.*

`check_issues_page.R` uses `uploaded` only as a liveness flag, to decide whether a note is DUE or merely PENDING. It could ask Redivis instead and the column would stop being able to lie. That would also have caught `himmelstein-admc_raw-2025` reading as PENDING while live, and the `batch_015` set reading as unpublished for the hours between upload and release.

Two things to weigh against it: the check currently runs with no credentials and no network, which is a real virtue; and `uploaded` also carries *when*, which Redivis cannot reconstruct after the fact. A middle option is to keep the column as the record of when, and have the checker warn when a blank `uploaded` names a table that is live.

## Also found

`ALSECYPIAMH_WU_2022_PHQ` has a provenance row in **both** `batch_004` and `batch_012`. One table, two batch records, neither stamped. Worth deciding which is authoritative before either is filled in.

Related: #1777 (where the disclosure reconciliation lives), #1820 (stamps the batch_015 eleven), `ARCHITECTURE.md` §4 on drafts vs releases.

https://claude.ai/code/session_01N5VsAkTyiVWtFij13gVYGa

ben-domingue:
Closes #1828.

`uploaded` was doing two jobs. One of them — *is this table live?* — Redivis
already answers, and the hand-kept copy was wrong in both directions:
`himmelstein-admc_raw-2025` read as PENDING while it was live, and 16 rows named
a live table with the cell blank. The other job — *when did it go up?* — Redivis
cannot answer at all, because opening a draft copies the dataset and resets every
per-table timestamp. So the column keeps the second job and loses the first.

## What changed

- **`itemtext/refresh_live_tables.py`** — writes `itemtext/live_tables.csv`, one
  row per item-text table with `status` of `published` or `draft`. Two API calls
  per shard, no exports.
- **`check_issues_page.R`** — takes liveness from that file. With no snapshot it
  still runs, falls back to `uploaded`, and says so in the header, so the
  credential-free/offline property the issue asked to preserve is intact.
- **Draft is its own state.** 12 tables are on the issues page *and* in the draft
  right now; they are neither live nor missing. They report as STAGED rather than
  as a page describing a table nobody can fetch.
- **Two new non-gating reports.** STAMP OWED (live, no date anywhere) and GONE
  (stamped, not live). Printed every run so they cannot accumulate the way #1828's
  sixteen did, but they do not change the exit status — they are bookkeeping, not
  an unwritten disclosure.
- **`check_provenance.R`** — reports one table with two provenance rows making
  different claims, which is the issue's "Also found". Five cases; each superseded
  row now names the record that holds, and the check goes quiet once it does.
- **Tests + CI.** `itemtext/tests/test_live_tables.py` pins the file contract
  between the Python writer and the R reader. Offline, so it joins `tests.yml`.

## What it found on the first run

- **GONE: 27 tables.** All explained — the PROMIS seven (#1927 era), the
  WHOQOL/TAS-20/DSES withdrawals, the wording_rights thirteen (#1955), and
  `alomari_2025_student_questionnaire`, whose response table also left the corpus.
  No upload silently failed.
- **STAMP OWED: none**, once `unrecorded` is read as the deliberate stamp it is.
- **DUE grew by 4** — `ALSECYPIAMH_WU_2022_PHQ`, `twod_rotation_mather2023`,
  `himmelstein-admc_raw-2025`, `enkavi_2019_simon`. These were live all along and
  were hiding in PENDING behind a blank cell. They owe issues-page entries; that
  is page-writing work, not part of this PR.

## On the duplicate

`ALSECYPIAMH_WU_2022_PHQ` is recorded in both `batch_004` (`paper_order`, held on
rights) and `batch_012` (`paper_explicit`, hold released 2026-08-26 per #1643).
On the files' own evidence batch_012 supersedes batch_004, so the batch_004 note
now says so rather than either row being deleted — the earlier record is still
the true account of what was known then.

https://claude.ai/code/session_01FpYsJsd2SHkwd3uqAcA4sz


ben-domingue:
Rebased onto main (through #2034) — the conflicts were with #2026, which added a `translation_source` column to `batch_028`/`batch_029`, and with #2034's withdrawal check in `check_provenance.R`. Both taken as-is, with this branch's changes reapplied on top. `tests` and `contract` pass.

#2034 also improved one thing here. `GONE` was printing 27 tables, 26 of which are deliberate withdrawals — a list that long is one nobody reads, and the single entry that means something was hiding in it. It now counts the recorded withdrawals and lists only the remainder.

Recognising them takes reading two fields, because the two rounds wrote it differently: `tools/withdraw_wording_rights.py` rewrites the `public_note` to open "IRW does not offer item text for" (the signal #2034 reads), while the PROMIS round wrote `WITHDRAWN` at the head of the private `note` and left `public_note` empty. A withdrawal recorded in the wrong field is still a withdrawal.

That leaves one: `alomari_2025_student_questionnaire`. Its response table left the corpus too, so the item text going with it is consistent — but nothing anywhere says so.

Worth noting separately: the seven PROMIS rows have no `public_note` at all, so the public "IRW does not offer item text for this instrument" statement that every wording_rights withdrawal carries was never written for PROMIS. Not this PR's business, but it is the same shape as withdrawals not reaching the site.

PR #2041, opened 2026-09-07
itemtext: check the 24 unverified source-variable-labels claims (#1745)

### LINKED ISSUE #1745: Itemtext: 23 mapping rows claim source variable labels without ever being checked
## Summary

23 rows in `itemtext/mapping_verification.csv` record `route=source_variable_labels` with `status=NOT_NEEDED` — i.e. the item-to-text mapping is asserted to need no verification because the source file's own variable labels tie code to text. That is the strongest basis there is **when the claim is true**. Nothing has ever checked that it is.

We now know of one case where it was false for eight months.

## The case that motivates this

`abouhashish_2025_chatgpt_attitudes` (found under #1645, corrected in #1737) carried:

```
data_labels / source_variable_labels / NOT_NEEDED
"Item code is tied to its text by the source data file's own variable labels /
 column headers, so the mapping is authoritative at the source. Numeric
 verification would add nothing."
```

Its item codes are `item_01`..`item_40` — **positional indices assigned by the processing script, not source column names.** There are no variable labels tying those codes to anything. The row claimed exemption from verification on a basis that did not exist, and positional assignment is the pattern the skill flags as highest-risk.

It turned out to be correct — all 40 shipped texts were diffed against the source `.xlsx` header at each position and the alignment holds — but that check had never been run, and a one-position shift would have gone unnoticed indefinitely. This is the `gilbert_meta_35` failure mode (#1615: 7 of 8 items carrying the next item's text, live for months, both gates passing) with the additional problem that the record said no check was needed.

## Why the gates cannot catch it

`audit_batch.R` and `diff_itemtext.R` compare *sets*. A shifted mapping has an identical item set, an identical resp set, and individually plausible text on every row. Both gates PASS. `fixes/HANDOFF.md` states this outright and the addy SDQ_7 defect (#1738) confirmed it again on the option axis.

## The check

A `data_labels` / `source_variable_labels` claim is **falsified by item codes that look like positional indices rather than source column names**. That is cheap and mostly mechanical:

1. Pull item codes for each of the 23 tables (`irw::irw_itemtext(<table>)`).
2. Flag any table whose codes are index-shaped — `item_01..item_NN`, `q1..qN`, `V1..VN`, or any dense integer sequence — rather than source-column-shaped.
3. For each flagged table, open the source and verify the mapping properly, then rewrite the row to say what was actually checked (as #1737 did for `abouhashish`).
4. Rows that survive keep `NOT_NEEDED`, but with evidence naming the labels rather than the boilerplate.

Note the two rows still carrying the verbatim "Numeric verification would add nothing" boilerplate — `16_personalityfactors` (batch_001) and `albeitawi_2025_preceptor_needs` (batch_002) — are a starting point but not the population. The boilerplate string is not the tell; the mismatch between claimed basis and actual code shape is.

## Scope

23 tables to screen, an unknown and probably small subset to verify by hand. Bounded, and it retires a category of silent risk across the whole published item text corpus rather than one table.

Related: #1645, #1737, #1738, #1615, #1736. Rolls up to #1709 (item 7, scale item text) — this is about trusting what is already shipped before multiplying it.

ben-domingue:
Closes #1745.

## What the rows were claiming

`mapping_basis=data_labels` + `status=NOT_NEEDED` asserts that the source file's own variable labels tie item code to item text, so no numeric verification is owed. Nothing had ever checked that the source carries such labels, or that the IRW item code is the source column name at all. Both gates compare sets; `lint_verification.R` exempts this combination by construction. Confirming that: the lint output is byte-identical before and after this change, including on the two rows that turned out to be false.

## Population

29 rows on `route=source_variable_labels` with `status=NOT_NEEDED` — the issue's 23, one added since, and five from batches 039/040 (including three `gilbert_meta_*`). All screened; each then verified against its actual source. 31 rows rewritten.

## What the check actually was

The issue proposed screening on item-code *shape* (`item_01..item_NN` is index-shaped, so suspicious). That screen is weak in both directions: `UWES_1..UWES_17` is index-shaped and is also the literal source column name, while `busch_2023_stigma`'s codes look like content words and its source has no labels at all. The decisive test is what the processing script does with the header — whether `item` is the source column name or a position. That is mechanical and offline.

## Results

| | tables |
|---|---|
| Claim true, now evidenced | 13 |
| Claim sound, wrongly described | 6 |
| Claim false | 2 |
| Nothing shipped to verify | 3 |
| Narrowed to PARTIAL | 1 |
| Newly found by the screen, resolved | 2 |

**True and now evidenced (13).** Shipped text diffed line by line against the source's own labels: `an_2020_efl_self_regulated` 49/49 (the labels are GB2312 and come back mojibaked from pyreadstat — decode as GB18030 first; note the source renumbers from 1 at Q27 and again at Q34, and the shipped text follows the column name rather than that numbering, which is what a slip would have broken), `baka2023_{jcs,olbi,uwes}` 46/46, `arnulf_2022_general_knowledge` 12/12, `avilesgonzalez2019_ces` 30/30, `butt_2022_*` 8/8, `brederecke_2020_phq4` 4/4, `anjum_2022_gad7` 6/6. For `albeitawi_2025_preceptor_needs` and `bartoli_2022_badge_notifications` the .xlsx headers *are* the item text, so code and text are the same object and no mapping exists to be wrong.

**Sound but wrongly described (6).** `audretsch` and `emidy2024_fevs` are Stata, whose 80-char variable-label cap means the label pins the mapping but cannot carry the wording — for audretsch the label is an exact leading prefix of the shipped text for all 15 items; for FEVS all 93 reconcile once the questionnaire's shared-stem `…` convention is matched against the spelled-out label. `machivallianism_test_main` (20/20), `autonomysupport_mokken` (7/7) and `sapa_personality` take their text from a separate codebook, an R package help page, and an item lookup respectively — not from data labels at all. For sapa, `psychTools::spi.dictionary` independently covers 135 of the 696 codes and 134 agree.

**False (2).**

- `16_personalityfactors` — `data/16_personalityfactors.R` lowercases the headers, drops six covariate columns, then overwrites `item` with `row_number()`. The codes `1..163` are positional indices into the surviving column order — the #1737 pattern — and the openpsychometrics source is a bare CSV with no labels, its wording in a separate `codebook.html`. Verified by hand anyway: reconstructing the column order gives **162/163 exact**, and **0 matches under either a +1 or −1 shift**, so the alignment is pinned rather than merely plausible. One gap: item 163 is source column `p10`, which the codebook does not document (its entries stop at `P9`), so the shipped text there — "I have a good word for everyone" — has an origin outside the deposit and is unverified.
- `busch_2023_stigma` — also a bare CSV with no labels, so the exemption never applied. It cannot be checked per item because it is not per-item text: 6 of the 8 shipped strings are the identical stem "I most closely identify as", with the content in the response options. That is already described on the item-text issues page; this only stops the ledger claiming a basis the source does not provide. → `NO_ROUTE`.

**Nothing shipped (3).** `bakker_2020_pss10` and `close_relationships` are `blocked` in `queue_state.csv`, `garciabatista_2021_erq` is `failed`. All three were carrying a verification exemption for text that is not live; `garciabatista`'s row even carried detailed evidence for it. → `NO_ROUTE`.

**Narrowed (1).** `chen_2022_sasc` → `PARTIAL`. Its codes are the source .sav's own column names, so the code→column half is safe by construction, but the text is the published English SAS-C list. The residual assumption is that source column `SASC`*n* is published SAS-C item *n* — unverified here (the Science Data Bank deposit was not retrievable this pass) and exactly the assumption that failed in `buzgova_2023_rses`.

## The re-runnable half

`itemtext/check_label_claims.py` reads each claiming row's processing script and flags positional `item` assignment (tier 1) or a derived-but-reversible code (tier 2). Matching is per file, not per block, so a flag means "verify by hand", never "defect".

Run against the pre-fix ledger it rediscovers all three hand-found cases *plus two more* outside the issue's population, both resolved here: `psychoneurotic_inventory` (verified 115/116 against the WPI codebook — the one diff is an editorial gloss on a source typo) and `machivallianism_test_vcl` (per-file false positive; that table is built by `pivot_longer(starts_with("VCL"))` and keeps its names). It exits clean on the post-fix ledger.

## Needs a decision, not fixed here

- **13 of 696 `sapa_personality` items ship SAPA's abbreviated bracketed form** rather than the wording — e.g. `q_1483` ships `"[Forget to put things back]."` where the canonical is `"Often forget to put things back in their proper place."`. Faithful to the source lookup, but abbreviated. Recorded in the row; worth its own issue if we want them expanded.
- **The status vocabulary has no value for "no item text is live."** `lint_verification.R` allows only VERIFIED/PARTIAL/NO_ROUTE/NOT_NEEDED, so the three blocked/failed tables are parked on `NO_ROUTE`, which understates why. A fifth value would be clearer — deliberately not invented here.
- **The lint cannot see this defect class**, by design. Teaching it to run `check_label_claims.py` over `NOT_NEEDED`/`data_labels` rows would close the gap permanently; not wired in as part of this change.

https://claude.ai/code/session_01RoLJecpfggFyotzZnfRa3Z


ben-domingue:
I screened #1745 in parallel (my comment there crossed with yours by two minutes — sorry for the duplication). We agree on the method: the decisive test is what the processing script does with the header, not the code's shape. Same population definition too — your script and my screen both key on `mapping_basis=data_labels` + `status=NOT_NEEDED` = 100 rows.

I ran `check_label_claims.py` from this branch against the pre-fix ledger. **It misses five positional tables, in two distinct ways.** All five verify clean by hand, so there is no data defect — but all five carry a false evidence string, and the screen passes them.

## Gap 1 — Tier 1 is R-idiom-shaped; Python positional assignment slips through

Two tables have a script, are positional, and get neither a `[FLAG]` nor a `[NOTE]`:

| table | the line | why `POSITIONAL` misses it |
|---|---|---|
| `dong_2025_teacher_leadership` | `item_names = {c: f"item_{i+1}" for i, c in enumerate(item_cols)}` | dict comprehension; not `paste0("item_"`, not `["item"] = range` |
| `dou2025_area` | `long["item"] = [f"{prefix}_{i+1}" for i in ...]` | list comprehension, and the prefix is a variable |

The wider pattern is that the newer `automated_finding` Python scripts build the code with `enumerate` + an f-string over a positional column slice, which none of the seven Tier-1 patterns match. `dou2025_area` is the sharper example: `item_cols = list(raw.columns[27:39])` is a bare positional slice into a sheet that has `Unnamed` gap columns between instruments, which is about as exposed as this gets.

Suggested additions, in the spirit of "add to the list rather than loosening it":

```python
(re.compile(r"""f["'][^"']*_\{\s*i\s*\+\s*1\s*\}"""), 'f"..._{i+1}" over enumerate'),
(re.compile(r"""\[\s*["']item["']\s*\]\s*=\s*\[\s*f["']"""), 'item = [f"..." for ...]'),
(re.compile(r"\.columns\[\s*\d+\s*:\s*\d+\s*\]"), "positional column slice"),
(re.compile(r"for\s+\w+\s*,\s*\w+\s+in\s+enumerate\([^)]*\)\s*,\s*start\s*=\s*1"), "enumerate(..., start=1) rank"),
```

## Gap 2 — `scripts_for()` misses multi-table scripts that build the table name

Four tables land in `[SKIP] no processing script found under data/`:

```
conspiracy_asd__cognitive_flexibility
conspiracy_asd__conspiracy_gcbs
conspiracy_asd__schizotypy
dpt_noncog__emotional_intelligence
```

They all have scripts — `data/conspiracy_belief_schizotypy_asd.py` and `data/dpt_noncognitive_traits.py`. `scripts_for()` greps for the literal table name, and these scripts never contain it: the name is assembled from a `prefix` variable (`wide[f"{prefix}_{rank}"]`, `f"dpt_noncog__{scale_name}.csv"`). It is a silent miss, and `conspiracy_asd__*` is Tier 1 material — `range(32,52)` → `cfi_{rank}`.

A cheap fix is to fall back to progressively shorter prefixes of the table name on `__` and `_` boundaries before giving up, and to report a prefix match distinctly from a literal one. `SKIP` counts as a pass today, which is the wrong default for a gate — worth exiting non-zero on unresolved skips, or at least separating "no script" from "not screened".

## Hand-verification of all five, so they need no re-work

I diffed each against its source. Mapping is correct everywhere; only the evidence strings are wrong.

- **`dong_2025_teacher_leadership` — 16/16 verbatim** vs the PLOS S1 `.xlsx` headers. The `^\d+\.` columns are in numeric order, so `enumerate` order coincides with the numbering.
- **`dou2025_area` — 12/12 verbatim.** Slice `[27:39]` lands exactly on source questions 25–36 and correctly skips `Unnamed: 26` / `Unnamed: 39`.
- **`conspiracy_asd__cognitive_flexibility` / `conspiracy_gcbs` / `schizotypy` — 72/72 aligned** against the source's header row 1 (the Qualtrics `._1`.. suffixes confirm rank order).
- **`dpt_noncog__emotional_intelligence` is a genuine exemption.** Its source workbook's header row 2 reads `Schutte_1, Schutte_2, Schutte_7…`; the script's hardcoded `Q20_n → schutte_n` map is a lowercased copy of that row, and `provenance.csv` already says so. This one is correctly described — it just should not have been a `SKIP`.

~~The three `conspiracy_asd__*` rows do need rewriting for a different reason, filed as #2042.~~ **Retracted** — #2042 is closed as wrong. I had diffed the shipped text against the coded workbook's header row, but it comes from the deposit's `Plain text version of survey.docx`, which carries the uncorrupted wording and the terminal periods. It is a literal transcription, and `batch_023/provenance.csv` documents the source choice and the header's `n`→`2` corruption in detail. The five tables above still need their evidence strings rewritten; the `conspiracy_asd__*` provenance notes are already accurate and are a good model for the rest.

## One more, on your "evidence is a copy job" point

Of the 100 population rows, 74 have a `provenance.csv` note and **70 of those name a concrete mechanism** ("header rows 1-4", "the .sav's own variable labels", "row 2 the original scale item number"). For **43 the `mapping_verification` evidence is boilerplate instead.** The evidence #1745 asks someone to go write already exists one file over for most of them — which also suggests the durable fix is to source `evidence` from provenance rather than let an agent re-assert it.

ben-domingue:
Both gaps fixed in cf66d83. Your five are all caught now, and nothing previously caught is lost.

## Gap 1

I started from your suggested patterns but had to narrow them: the first cut flagged 60+ rows. The over-match is that most of these scripts contain a line like

```python
ITEM_COLS = [f"Q{i}" for i in range(1, 51)]
```

which is the *opposite* of the defect — it generates source column names in order to select them, so the strings have to equal real headers or the selection raises. `f"...{i+1}"` on its own can't separate that from `dong`'s rename. The tell is **`enumerate` over the columns** (position → new name) or an assignment into `item`. Three patterns:

- dict-comprehension rename over `enumerate` — the loop index is bound with a **backreference**, so the f-string has to be numbered by the same variable `enumerate` supplies (catches `dong`)
- `["item"] = [f"..."` (catches `dou2025_area`)
- a new column named by a loop rank, `wide[f"{prefix}_{rank}"] =` (catches `conspiracy_asd__*`)

I dropped your `enumerate(..., start=1)` suggestion — pattern 3 already catches `conspiracy_asd__*`, and standalone it hits scripts that enumerate for unrelated reasons. `.columns[27:39]` went in as a **Tier-2 NOTE** rather than Tier 1: it isn't a code assignment, but you're right that the column *set* can shift under it.

Worth knowing: that `enumerate` rename idiom is not rare. It appears in ~20 scripts (`sas_{i}`, `ibdq_{i}`, `{prefix}{i+1}`) and I spot-checked three — all genuinely positional. This is a bigger population than the five.

## Gap 2

Prefix fallback on `__` then `_`, reported distinctly from a literal match. Both your cases resolve at the first step: `conspiracy_asd` and `dpt_noncog` each match exactly one script.

Two additions beyond what you suggested:

- **An exact filename match now wins over a content match.** Without it the new patterns produced a cross-file false positive: `dasilva_2019_crt.py` renames `CRT_*` positionally and mentions `dasilva_2019_hexaco24` in a "companion to" comment, so the hexaco row got flagged on a script that doesn't build it.
- **A prefix that matches more than 4 scripts is noise, not a resolution.** `gilbert_meta` hits 14. Those rows stay unresolved, and the message now says which prefix and how many, so it points somewhere instead of just giving up.

`SKIP` is gone — unresolved rows fail the run, per your "wrong default for a gate". That currently leaves 4: `gilbert_meta_16/27/29` and `gad_BrummerHoffman_2021`.

`dpt_noncog__emotional_intelligence` now resolves and produces neither a flag nor a skip, which is the right answer for a hardcoded `Q20_n → schutte_n` map.

## Not done

The screen now exits 1 on this branch, correctly: `dong`, `dou2025_area` and the three `conspiracy_asd__*` still carry the false evidence strings you identified. I have your verification numbers and can write those five rows here — but three of them are `conspiracy_asd__*`, which is #2042's territory, and the verification is yours. Say the word and I'll do it, or leave them to you.

Your provenance point is the more interesting one: if 70 of 74 rows already name a concrete mechanism one file over, sourcing `evidence` from `provenance.csv` rather than letting an agent re-assert it would retire this whole class rather than this instance of it. Happy to take that as a separate change.

PR #2091, opened 2026-09-08
fix(enem): filter absentees by INEP's presence flag, drop annulled items (#1942)

mateusmazza:
Fixes findings 1 and 3 of #1942, and adds the post-scoring guard that would have caught both. **All 13 years rebuilt** and diffed against the pre-fix build of the same code base.

Finding 2 is **not** in this commit: it was already fixed by `4478e89` and is simply unpublished — see the section at the bottom.

## Finding 1 — the cause is the absence *encoding*, not the scoring rule

I did **not** take the proposed `case_when(raw %in% c(".","*") ~ NA_real_)`. Scoring `.` and `*` as 0 is INEP's own scoring and this keeps it. The 45% has a different cause:

| years | absent candidate's `TX_RESPOSTAS` | effect |
|---|---|---|
| **2013, 2014 only** | **all dots** (100% of absentees, all four areas, both years) | not NA → survives `drop_na()` → every `.` scores 0 |
| **2015–2025** | **empty string** | vroom reads `""` as NA → `drop_na()` already removed them |

The absence *rates* are similar (31% in 2013, 22% in 2019). Only the encoding differs, so **the blast radius is 8 tables, not 52.**

The fix uses INEP's own per-area flag `TP_PRESENCA_{CN,CH,LC,MT}` (0 absent / 1 present / 2 eliminated), keeping only presence 1 in all four areas. Omissions by candidates who *did* sit are only 0.2–0.5% of cells once absentees are gone, and `resp_raw` ships alongside `resp` for anyone who wants them missing.

**The frame was audited on the full file for all 13 years before any code changed**, because the load-bearing claim is that 2015–2025 are untouched:

- **11 of 13 years: the presence-1 set is *exactly* the set `drop_na()` already kept.**
- 2013: 2,165,629 rows drop, **0 added**. 2014: 2,774,337 drop, 2 added — and those 2 are candidates INEP marks present but ships empty strings for, so `drop_na()` still removes them.
- Every year's regular pool stays ≥ 1M (smallest 2.14M, 2021).

Confirmed in the rebuild: **id sets identical for all 44 tables in 2015–2025**, differing only for 2013×4 and 2014×4.

Share of candidates scoring exactly zero:

| table | old → new | median item p |
|---|---|---|
| `enem_2013_1mil_lc` | 44.8% → **0.07%** | 0.216 → 0.387 |
| `enem_2013_1mil_ch` | 43.1% → **0.01%** | 0.201 → 0.355 |
| `enem_2013_1mil_cn` | 43.1% → **0.03%** | 0.121 → 0.214 |
| `enem_2013_1mil_mt` | 44.8% → **0.11%** | 0.127 → 0.230 |
| `enem_2014_1mil_lc` | 46.8% → **0.09%** | 0.198 → 0.371 |
| `enem_2014_1mil_ch` | 45.4% → **0.01%** | 0.214 → 0.394 |
| `enem_2014_1mil_cn` | 45.4% → **0.03%** | 0.143 → 0.262 |
| `enem_2014_1mil_mt` | 46.8% → **0.14%** | 0.115 → 0.216 |

## Annulled items dropped

INEP records `TX_GABARITO = "X"` for an item it annulled after the exam. No response letter can equal `X`, so it scores 0 for every candidate — a zero-variance column, useless for IRT and indistinguishable downstream from a keying failure. Ten sit in the standard item sets and are now dropped before the item set is built:

`2018 MT 30294` · `2020 CN 32352, MT 44322` · `2021 MT 117674` · `2022 MT 39443` · `2023 MT 14887` · `2024 CN 61076` · `2025 CN 141557, CN 141774, MT 31350`

Nine tables lose one item; 2025 CN loses two.

## Finding 3 — @ben-domingue's checks, with two changes

`data/enem_checks.R` and `data/tests/test_enem_checks.R` are from `fix/1942-enem-blank-scoring`, with two changes, each commented at the point it occurs:

**1. The `zero_variance` `stop()` would have halted 7 of 13 builds.** The ten annulled items above each score 0 for every candidate. As written the build stops for 2018 and 2020–2025, and the message — *"an item nobody gets right is a keying failure, not a hard item"* — would be wrong: it's neither. Dropping them upstream is what makes the hard stop sound, so `p == 0` really does mean a keying failure.

**2. `share_all_wrong` was structurally blind to every LC table.** `complete <- answered == ncol(w)` is never true on LC, where the two elective language blocks are mutually exclusive and every candidate answers 45 of 50 columns — so `all_wrong` was forced to 0, silencing finding 1 exactly where it was reported (`enem_2013_1mil_lc`, 44.7%). Now keyed off the modal answered-count; identical to `ncol(w)` for CH/CN/MT. Two LC fixtures added to the test file. The rebuilt 2013 LC reports **0.2%**, a number the original would have printed as `0.0%`.

Wired into the existing `tests.yml` — base R, no microdata, no network. **The false-positive test holds on real data:** `enem_2013_1mil_cn` has 47% of items below the 1/5 chance floor with median item-rest 0.092 and was *not* failed. Difficulty alone never fails a build.

## Safety

**`resp` is unchanged on every `id`+`item` cell shared with the previous build, across all 52 tables.** The fix changes *which candidates are sampled*, never how a response is scored. That's the assertion I'd most want checked.

No build was stopped by the new checks, so across all 52 tables there is no zero-variance item and no form lacking item-total structure.

The 13 scripts remain **byte-identical apart from `year` and their two prova-code lines**, so the patch is verifiable the way #1942 asked. No refactor, for the reason given there.

## Finding 2 is not here

The LC 1:1-split bug was already fixed by `4478e89` and is simply unpublished. It affects **six** live LC tables, not one — 2015, 2016, 2018, 2019, 2020, 2021, every year where `TX_RESPOSTAS_LC` is 50 wide over a 45-position span. (2017 is genuinely 50/50; 2013/2014 are 45/45, which is why 2013 LC looks healthy.) The rebuilt `enem_2019_1mil_lc` gives median p **0.392** and median item-rest **0.330**, against 0.16 and 0.010 on the live table. The positional hypothesis in the issue isn't it: the LC electives *are* at positions 1–5 exactly where the branch assumes.

## Needs a separate step

**The 52 rebuilt tables need re-uploading** — this PR is only the code. That also makes `resp_raw` public for the first time (so the Dict/metadata needs the extra column), and the **nominal set must be regenerated**, since `data/nominal/enem.R` derives it from the published tables via `irw_fetch`.

Incidental: the 2020 absence rate is a real outlier — 44.8% present against 66–73% every other year. The COVID administration, now visible in the build log.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -130,6 +130,16 @@ jobs:
             --rpkg  _deps/Rpkg/R/redivis-config.R \
             --pypkg _deps/Python-pkg/src/irw/config.py
 
+      # The ENEM post-scoring checks (#1942). Base R only -- no microdata, no
+      # network -- so the thresholds that decide whether a 45-million-row table
+      # is publishable are exercised on every PR rather than only when someone
+      # has 13 years of INEP files on disk. The test that matters is the
+      # false-positive one: a legitimately HARD form must still pass, because
+      # difficulty alone must never fail a build.
+      - name: enem scoring checks
+        working-directory: data
+        run: Rscript tests/test_enem_checks.R
+
   # A PR to data/ contains a conversion SCRIPT, never a table --
   # automated_finding/irw_output/ is gitignored, so no output is ever in the
   # diff. This checks the script's contract; the data is checked by
--- a/data/enem_2013.R
+++ b/data/enem_2013.R
@@ -8,6 +8,20 @@ regular_prova_codes  <- c(167, 170, 171, 174, 176, 177, 178, 180, 181, 182, 189,
 # INEP response codes: A-E, "." blank, "*" double mark
 ALPHABET <- c("A", "B", "C", "D", "E", ".", "*")
 
+# the five that are actual answers; anything else in TX_GABARITO is not scorable
+ANSWER_KEYS <- c("A", "B", "C", "D", "E")
+
+# INEP's own per-area presence flag: 0 = absent, 1 = present, 2 = eliminated
+PRESENCE_COLS <- paste0("TP_PRESENCA_", c("CN", "CH", "LC", "MT"))
+
+# post-scoring sanity checks (#1942 finding 3). Resolved relative to THIS file
+# because the year scripts run with the microdata work dir as cwd, not data/.
+.enem_dir <- {
+  a <- grep("^--file=", commandArgs(FALSE), value = TRUE)
+  if (length(a)) dirname(sub("^--file=", "", a[1])) else "."
+}
+source(file.path(.enem_dir, "enem_checks.R"))
+
 # some years ship PARTICIPANTES + RESULTADOS instead of one MICRODADOS file
 find_ci <- function(dir, pattern) {
   hits <- list.files(dir, pattern = pattern, ignore.case = TRUE, full.names = TRUE)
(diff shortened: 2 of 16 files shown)


ben-domingue:
@mateusmazza and can i merge this do you think?

mateusmazza:
yes, should be all good.

Issue #2029, opened 2026-09-07
Corpus trust: 9.95M published rows carry resp = the literal string "NA"


ben-domingue:
Found while fixing `5personalityfactors` for #1842, then measured across the whole corpus. Every number below is from an aggregate query against the **live released** tables, 2026-09-06.

## The measurement

Two passes over all **4,238** response tables in the six `item_response_warehouse` shards:

1. read each table's declared `resp` type — **300 tables (7.1%) declare `resp` as a string**;
2. for those 300, count values that will not cast to a number.

**In 299 of the 300, the only non-numeric value is the literal two-character string `NA`.** Confirmed at byte level:

```
resp  len          n
  NA    2  3,715,652     <- dscore_bayley_weber_2019
   1    1    261,272
   0    1    162,692
```

| | |
|---|---|
| tables affected | **299** |
| rows where `resp` = `"NA"` | **9,951,293** |
| rows in those tables | 216,715,710 |
| tables >50% `"NA"` | 18 |
| tables >25% | 31 |
| tables >10% | 63 |

The 300th, `parenting_anunciacao_2025_material_rewards`, is a different thing entirely: 129 rows of untranslated Portuguese response labels (`Quase nunca`, `Raramente`, `Às vezes`, `Frequentemente`, `Muito frequentemente`). Real text, not a missing code — worth its own look.

## Worst affected

| table | rows | `"NA"` | share |
|---|---:|---:|---:|
| `dscore_bayley_weber_2019` | 4,139,616 | 3,715,652 | **89.8%** |
| `dscore_griffiths_weber_2019` | 1,508,416 | 1,373,321 | **91.0%** |
| `dscore_dutch_weber_2019` | 953,154 | 787,033 | 82.6% |
| `quopl2_forster_2021_elfe` | 546,356 | 509,272 | **93.2%** |
| `Veterans_Affairs_SSVF_Survey_2018-20` | 1,475,369 | 476,509 | 32.3% |
| `western_reserve_project` | 1,445,422 | 378,887 | 26.2% |
| `geography` | 10,087,305 | 281,706 | 2.8% |
| `dscore_asq_weber_2019` | 191,406 | 172,867 | 90.3% |
| `face_memory_test` | 309,400 | 157,566 | 50.9% |
| `issueirt_votes_shin_2024` | 303,921 | 130,615 | 43.0% |

Full list: `irw_validate/results/resp_string_na_2026-09-06.csv` (PR to follow).

## Why it matters

[...]

PR #2104, opened 2026-09-08
literal n/a patch

stacycremer:
This fixes the validator blind spot described in #2029. Pandas currently converts literal response values such as "NA" into missing values before the numeric check runs, allowing affected files to pass validation.

Changes:

Preserve literal response tokens when reading CSV, TSV, and TXT files under the upload and legacy profiles.
Flag every nonnumeric response, including values below the existing 1% tolerance, with counts and examples.
Continue treating genuinely empty cells as missing.
Prevent mixed numeric/text responses from crashing the response-scale check.

All 72 local validator tests passed, including 15 new regression tests covering file parsing, missing values, rare invalid responses, CLI behavior, and the uploader’s local validation step. Hosted CI has not run.

Two choices warrant review: removing the 1% tolerance and adding a second read of the response column. Large-file performance has not been benchmarked, and the existing 512 MiB validation limit remains.

This does not modify historical tables or published response counts. The broader cleanup in #2029 remains outstanding.

Related: #2029.


stacycremer:
Removed automated_finding/irw_claude_assist.py from this branch. It was uploaded here by mistake.

This PR is now 4 files, +211 -6, all in irw_validate.

ben-domingue:
Thanks @stacycremer — this is the right fix in the right place, and the diagnosis is exactly correct. The reason `resp_numeric` never fired is that `pd.read_csv(..., sep=None, engine="python")` deletes the two characters the check exists to catch, so the gate was reporting pass on its own defect. Moving the fix to the *read* rather than the check is the correct call, and your note about the python parser applying NA filtering after `converters=` is right — a converter alone genuinely can't do it.

**One thing to fix before I merge:** the branch also adds `automated_finding/irw_claude_assist.py` (442 lines), which isn't related to #2029 and isn't mentioned in the description. Looks like it rode along with one of the "Add files via upload" commits. Could you drop it from this branch? Happy to look at it separately as its own PR — though heads up that its `DEFAULT_MODEL = "claude-sonnet-4-6"` isn't a real model id, so it would fail on the first API call.

**Two things that are not your problem:**

- The red `contract` check is a CI artifact, not a defect in your PR. It ran with `base="origin/stacycremer-patch-1"` instead of `origin/main`, so the merge base is ancient and every script in `data/` got classified as "added by this PR". Your diff touches nothing in `data/`. `tests` passes. I'll fix the workflow separately.
- Removing the 1% tolerance: keeping it. You were right to flag it as a judgment call. The blast radius is small and points the right way — nothing in shards 3–6 carries the token, so the only thing that now blocks is a re-upload of one of the 300 legacy tables, which is exactly when we want to be stopped.

One note for the record, no change needed: the strictness lands on `validate_frame`'s upload/legacy gate, not on `run_qc` — which is what the `data/` scripts actually call, and which keeps the old tolerance (one of your tests pins that). So a processing script can still go green on a table the uploader then refuses. Your README paragraph documents this correctly; closing the gap is a separate piece of work.

Note also that this doesn't touch the ~9.94M published rows already carrying the token — that's the other half of #2029 and we're scoping it now.

stacycremer:
I believe the automated_finding/irw_claude_assist.py was already removed. Let me know if that isn't the case. It was added by mistake, no need to look at it separately.

ben-domingue:
Confirmed — you removed it, and my earlier comment was looking at a stale diff. `52aaedde` ("Remove irw_claude_assist.py, unrelated to #2029") landed at 18:12 UTC, about 13 minutes before I asked for it. Sorry for the noise.

For the record on why it looked otherwise: there are two branches with this name. The PR head is `stacycremer/irw:stacycremer-patch-1`, which is clean — `main...head` is 4 files, +211/-6, all under `irw_validate/`. But my own repo also has a `stacycremer-patch-1`, still parked at `d19f592d` from June, and relative to `main` that branch contains exactly one file: `automated_finding/irw_claude_assist.py`. So the path resolves under that ref while being absent from your PR. That stale branch is also what the `contract` check picked up as its base, which is where the spurious red came from.

I'll delete the leftover branch on my side. Nothing outstanding on yours — noted that the assist tool doesn't need a separate look. `mergeable: MERGEABLE`, base `main`; merging.

Issue #2114, opened 2026-09-08
li_2024_spa ships four reverse-keyed items un-reversed


ben-domingue:
Found while verifying #2113. `li_2024_spa` is Xu's (2003) Chinese 15-item Social Physique Anxiety Scale, and four of its items are reverse-keyed in the instrument but are shipped in raw form, so a naive total is wrong.

### Three independent signs, all pointing at the same four items

**1. Item-total correlations.** Corrected item-total r within the shipped table:

| item | r | | item | r |
|---|---|---|---|---|
| **spa_1** | **-0.004** | | spa_9 | +0.678 |
| **spa_3** | **+0.017** | | spa_8 | +0.657 |
| **spa_7** | **+0.035** | | spa_12 | +0.649 |
| **spa_14** | **+0.080** | | spa_6 | +0.647 |
| spa_11 | +0.223 | | others | +0.38 … +0.63 |

Four items sit at ~0 against ten in the +0.38–0.68 band.

**2. The paper quotes one of them, and it is plainly positively worded.** Li et al. 2024 ([10.7717/peerj.17910](https://doi.org/10.7717/peerj.17910)) illustrate the scale's dimensions with *"item 3, **'I feel comfortable even though other people's body size is superior to mine'**"*. Comfort is low anxiety — `spa_3` is reverse-keyed, and it is one of the four.

**3. Reversing exactly those four reproduces the paper's published statistics; the shipped scoring does not.**

| scoring | mean | SD | alpha |
|---|---|---|---|
| **published (Li et al., Table 1 + text)** | **43.21** | **10.52** | **0.877** |
| as shipped (no reversal) | 43.56 | 9.00 | 0.805 |
| **reverse {1, 3, 7, 14}** | **43.17** | **10.43** | **0.873** |
| reverse {1, 3, 7, 14, 11} | 44.21 | 10.69 | 0.883 |

All three published figures are matched by the four-item set and missed by the shipped one — note especially the SD, 10.43 vs a shipped 9.00 against a published 10.52. Adding `spa_11` (the ambiguous +0.223) moves every statistic further away, so the set is four, not five.

### Why it matters

`resp` direction is not harmonised across IRW, so un-reversed items are not by themselves a defect. What makes this one is that **the shipped table cannot reproduce its own paper's descriptives**: any user summing the 15 items gets SD 9.00 and alpha 0.805 where the study reports 10.52 and 0.877, with no note saying why. This is the same family as #2099 and #2100.

[...]

PR #2115, opened 2026-09-08
li_2024_bdyz is the ERQ expressive-suppression subscale, not a body-image scale (#2113)

### LINKED ISSUE #2113: li_2024_bdyz dictionary Description is wrong: it is the ERQ expressive-suppression subscale, not a body-image scale
Two IRW records disagree about what this table contains, and the dictionary is the one that is wrong.

- **Dictionary Description:** `4-item body-image scale subset (items 1/2/6/9 only, 1-7 Likert), same sample as li_2024_fa`
- **`itemtext/availability_audit_full.csv`:** *"BDYZ is the Expressive Suppression subscale derived from Gross & John's (2003) Emotion Regulation Questionnaire (ERQ)"* — correct.

### The paper says so outright

Li et al. 2024, PeerJ [10.7717/peerj.17910](https://doi.org/10.7717/peerj.17910) (CC BY 4.0, Europe PMC `PMC11332389`):

> The Expressive Suppression Scale is one of the subscales of the Emotion Regulation Strategies Scale, a scale that is widely used worldwide and has been licensed by the original authors for use in all Chinese populations (Gross & John, 2003; Zhao et al., 2015). It uses a 7-point Likert scale with scores ranging from 1 to 7 … **This subscale has four items** (e.g., item 9, "I don't express sadness and anger when I feel it") and the total score ranges from 4 to 28.

The study's body-image measure is the **separate** 15-item Social Physique Anxiety block, already shipped as `li_2024_spa`.

### Re-derived from the live table

Independently computed from `li_2024_bdyz`, against the paper's own Table 1:

| statistic | published | derived from IRW |
|---|---|---|
| Expressive suppression mean | 14.66 | **14.6603** |
| SD | 5.61 | **5.6147** |
| N | 1,151 | **1,151** |

The four columns sum to the paper's expressive-suppression total exactly. Cronbach's alpha 0.8336 on 4 items, 1–7 response range, 1,151 complete rows.

### Suggested replacement

```
4-item Expressive Suppression subscale of the ERQ (Gross & John, 2003; 1-7 Likert; deposit ships the items labelled 1/2/6/9), same sample as li_2024_fa
```

To be applied through `DESCRIPTION_OVERRIDES` in `metadata/dict_union.R`, not by hand-editing the sheet.

### One correction to the source note

`itemtext/itemtables/batch_071/notes.csv` says the derivation reproduces "its Table 2 correlation with social physique anxiety (0.219) exactly". It does not: r(ES, SPA) computed from the live tables is **0.317** with SPA scored as shipped, and **0.248** with SPA's reverse items reversed. Neither is 0.219. The mean/SD/alpha match is exact and is what actually establishes the identification; the correlation claim should not be relied on. That mismatch is a symptom of a separate defect in `li_2024_spa`, filed separately.

Found during item-text extraction (batch_071, #1709); not previously filed.

ben-domingue:
Closes #2113.

Two IRW records disagreed. The dictionary said *"4-item body-image scale subset (items 1/2/6/9 only, 1-7 Likert)"*; `itemtext/availability_audit_full.csv` already had it right as the ERQ expressive-suppression subscale.

**The paper settles it without inference.** Li et al. 2024, PeerJ [10.7717/peerj.17910](https://doi.org/10.7717/peerj.17910), CC BY 4.0:

> The Expressive Suppression Scale is one of the subscales of the Emotion Regulation Strategies Scale … **This subscale has four items** (e.g., item 9, "I don't express sadness and anger when I feel it")

**Re-derived from the live table** against the paper's Table 1:

| statistic | published | derived |
|---|---|---|
| mean | 14.66 | 14.6603 |
| SD | 5.61 | 5.6147 |
| N | 1,151 | 1,151 |
| alpha | — | 0.8336 (k=4, 1–7) |

The study's body-image measure is the separate 15-item Social Physique Anxiety block, already shipped as `li_2024_spa` — so the old text pointed at a different scale in the same study.

**Applied through `DESCRIPTION_OVERRIDES`, not by hand-editing the sheet.** Verified the entry actually fires: run against the live sheet's current cell it reports `applied 1 of 6` and rewrites the Description, so it is not one of the inert entries the stale-override warning exists to catch. `metadata/tests/test_dict_union.R` passes.

**Found on the way, filed as #2114:** the source note claimed this derivation also reproduces the paper's Table 2 correlation with social physique anxiety (0.219) "exactly". It does not — r(ES, SPA) is 0.317 as shipped. Chasing that gap turned up a real defect in `li_2024_spa`: four reverse-keyed items ship un-reversed, so the table cannot reproduce its own paper's SD (9.00 vs a published 10.52) or alpha (0.805 vs 0.877). The mean/SD/N/alpha match above is exact and is what actually establishes this identification.

https://claude.ai/code/session_01DyMmEFLmpZaoybzpsfWGow

### DIFF
--- a/metadata/dict_union.R
+++ b/metadata/dict_union.R
@@ -645,6 +645,26 @@ refresh_biblio_from_dict <- function(biblio, dict, label = "core", log.file = NU
 ##should not fail because someone improved a sentence in the sheet.
 DESCRIPTION_OVERRIDES <- list(
     list(
+        table      = "li_2024_bdyz",
+        issue      = "#2113",
+        superseded = "4-item body-image scale subset (items 1/2/6/9 only, 1-7 Likert), same sample as li_2024_fa",
+        corrected  = paste("4-item Expressive Suppression subscale of the ERQ (Gross & John,",
+                           "2003; 1-7 Likert; deposit ships the items labelled 1/2/6/9),",
+                           "same sample as li_2024_fa"),
+        ##Not an inference from the code name: the paper says it outright --
+        ##"The Expressive Suppression Scale is one of the subscales of the
+        ##Emotion Regulation Strategies Scale ... This subscale has four items
+        ##(e.g., item 9, 'I don't express sadness and anger when I feel it')".
+        ##Re-derived from the live table against the paper's Table 1: mean
+        ##14.6603 vs published 14.66, SD 5.6147 vs 5.61, N 1,151 vs 1,151, alpha
+        ##0.8336 on 4 items over a 1-7 range. The study's body-image measure is
+        ##the separate 15-item Social Physique Anxiety block, already shipped as
+        ##li_2024_spa -- so the superseded text pointed at a different scale in
+        ##the same study. availability_audit_full.csv already had this right,
+        ##which is how two IRW records came to disagree.
+        why = "the paper names it the ERQ expressive-suppression subscale; mean/SD/N/alpha reproduce exactly"
+    ),
+    list(
         table      = "cdm_timss07",
         issue      = "#1898",
         superseded = "Subsample of TIMSS items from Australia with Q matrix",


ben-domingue:
Correction to one line in the description: the source note's claim that the derivation reproduces the paper's r = 0.219 "exactly" is right after all. It holds when `li_2024_spa` is scored with its six reverse-keyed items reversed (pinned in #2114); the 0.317 I quoted is SPA as shipped. A reader checking the claim against the published IRW table still gets 0.317, which is the practical point and is now filed as #2114. The identification fixed by this PR rests on the mean/SD/N/alpha match and is unaffected.

PR #2185, opened 2026-09-15
batch_204: 11 tables shipped, 1 blocked (#1945)

xingyi-zhang:
Eleven tables shipped, one blocked. Stacked on `xingyi/1945-batch203` (#2184) — merge that first and this diff becomes batch_204 alone.

## Shipped

| table | rows | items | status | what carries the mapping |
|---|---|---|---|---|
| `estcrm_selfeff` | 987 | 10 | VERIFIED | CRAN package column names; **3070/3070** cells reproduced |
| `wine_luckett2021` | 118 | 3 | VERIFIED | script's column rename; scale shape separates the intensity item |
| `FEDSP_Trzcinska_2023_MonKnow` | 4 | 2 | VERIFIED | `.sav` variable labels; deposit filter + dedup reproduced |
| `NAMPRB_Siwiak_2024_SSUB` | 125 | 25 | VERIFIED | numbered bilingual supplement; **40/40** vectors reproduced |
| `FIVPEI_Perrig_2023_AttDiff` | 147 | 21 | PARTIAL | block certain, within-block order assumed; **21/21** vectors |
| `FAD_fadplus_goto2021` | 270 | 54 | VERIFIED | supplement's 1-27 numbering; **54/54** vectors; factors from the authors' lavaan model |
| `gilbert_meta_112` | 23 | 8 | PARTIAL | `Item_1..8` columns; 3 of 8 items ship (parallel versions) |
| `chile_…_u` | 691 | 38 | VERIFIED | official codebook; **exact frequency reproduction** |
| `chile_…_ss` | 72 | 13 | VERIFIED | same |
| `chile_…_rr` | 110 | 33 | VERIFIED | same |
| `florida_twins_par` | 146 | 31 | VERIFIED | LDbase codebooks; **both composites recomputed per person, exactly** |

Eleven verify scripts, one per table, all `VERDICT: PASS`. Six are exact reproductions rather than arguments.

## Two things worth a look

**`florida_twins_par` verifies itself.** Two of its 31 codes are not items — `parstrict` and `parwarmth` are the codebook's sum composites, with the formulas printed. Since their inputs are in the same table, both were recomputed per person and matched exactly (**746/746** and **760/760**), reading no external file. That jointly tests the item mapping, the response-option order for every item in both blocks, and the recode direction at once.

It also recovered two behaviours the codebook does not document:

- `npar6` is **not** the reversal the codebook's own table prints for it. The data uses the raw curfew value with "As late as I want" folded to 0. The codebook flags this obliquely — *"Item 6 is coded differently than the original scale."* Found because the residual was a clean step function of `par6`, then confirmed by the exact match.
- `parwarmth`'s pair rule takes the mean of the **recoded** pair when both the father and mother item are answered, but that one's **raw** value when only one is. All 12 rows that failed a plain recoded-mean reading are half-missing pairs.

The two composites ship with a bracketed note in place of an administered stem and one row per observed value — they have to appear for the item sets to match, and the same treatment is already used for the chile numeric items.

**`NAMPRB`'s anchors do ship, and my first note was wrong to say they didn't.** I had read only the OSF deposit, whose IRT outputs label categories generically as `P1..P5`. The study also has a CC BY preprint (`10.31234/osf.io/7u98d`) stating the scale outright. The anchors are English while the items are Polish — deliberately: the administered Polish anchor wording is printed nowhere, so `option_text` takes the documented fallback rather than inventing it, disclosed in provenance so it reads as a rendering while `item_text` stays a transcription.

## Blocked

`gilbert_meta_113` — on data, not rights; the deposit is CC BY and prints the DNA test in full. Two blockers, neither fatal alone:

1. Both versions print Vraag 1-9 while the spreadsheet carries exactly `Item_1..Item_8`. One printed question is unscored or merged and nothing says which, so a drop at question *k* shifts every code from *k* on.
2. The two parallel versions diverge (identical only at Vraag 1, 4, 8) and the table records no version.

That is why 112 ships and 113 does not off the same spreadsheet and script: the Forces test has exactly eight questions, so blocker 1 never arises and blocker 2 still leaves three items attributable. Recoverable upstream, so logged in `pending_index_notes.csv` rather than excluded.

## Gates

`validate_items` (item + resp sets, all 11) · `normalize_nulls` · `check_provenance` · `check_label_claims` · `lint_verification` (11 rows, no problems) · `audit_batch` 5 PASS / 6 WARN, every WARN a documented case — the chile numeric items, the two florida composites, `gilbert_meta_112`'s five version-varying items, and NAMPRB's five collided retest codes, which is the defect itself.

`queue_state.csv`: the 11 tables → `done,batch_204`; `gilbert_meta_113` → `blocked,batch_204`.

## Data-side findings

Six defects in the **response** data surfaced across batches 203-204 — dropped form variables, a mis-thresholded item, 1,505 Excel formula artefacts, 310,922 `resp = NA` rows, a `gsub` that creates 5 spurious item codes, and 17 `id` values for 12 wine samples. None block item text; all are upstream of `itemtext/`. Being filed as a single separate issue.


ben-domingue:
@xingyi-zhang might want to check gilbert_meta_112__items and gilbert_meta_70__items

```
red_up -- 20 CSV file(s) under /tmp/ene/irw_itemtext_batches_203-204
checking 8 datasets for existing tables ...

target: datapages.irw_text_2:next   (item text (__items.csv) -- newest shard)

  NEW        FAD_fadplus_goto2021__items                         270 rows   13 cols  
  NEW        FEDSP_Trzcinska_2023_MonKnow__items                   4 rows   11 cols  
  NEW        FIVPEI_Perrig_2023_AttDiff__items                   147 rows   10 cols  
  NEW        NAMPRB_Siwiak_2024_SSUB__items                      125 rows   12 cols  
  NEW        chile_2023_social-welfare-survey_rr__items          110 rows   14 cols  
  NEW        chile_2023_social-welfare-survey_ss__items           72 rows   14 cols  
  NEW        chile_2023_social-welfare-survey_u__items           691 rows   14 cols  
  NEW        estcrm_selfeff__items                               987 rows   10 cols  
  NEW        florida_twins_par__items                            146 rows   10 cols  
  NEW        gilbert_meta_13__items                               74 rows   10 cols  
  NEW        gilbert_meta_14__items                              138 rows   10 cols  
  NEW        gilbert_meta_23__items                              117 rows   11 cols  
  NEW        gilbert_meta_30__items                               20 rows   10 cols  
  NEW        gilbert_meta_45__items                              214 rows   10 cols  
  NEW        gilbert_meta_68__items                               70 rows   11 cols  
  NEW        gilbert_meta_69__items                               60 rows   11 cols  
  NEW        gilbert_meta_71__items                               40 rows   13 cols  
  NEW        wine_luckett2021__items                             118 rows   10 cols  
  SKIP       gilbert_meta_112__items                              23 rows   13 cols  dup_item_resp: 5 duplicate item+resp rows with identical option text -- an item text table carries one row per response option, so a repeat is a doubled upload (#1810)
  SKIP       gilbert_meta_70__items                               90 rows   14 cols  dup_item_resp: 16 duplicate item+resp rows with identical option text -- an item text table carries one row per response option, so a repeat is a doubled upload (#1810)

  warnings (11)
    FAD_fadplus_goto2021__items                  name is not lowercase (breaks case-sensitive metadata joins)
    FAD_fadplus_goto2021__items                  name_charset: table name 'FAD_fadplus_goto2021' is not lowercase [a-z0-9_.] -- Redivis keeps the case but every client lowercases when joining, so a capitalised name silently drops out of case-sensitive joins.
    FEDSP_Trzcinska_2023_MonKnow__items          name is not lowercase (breaks case-sensitive metadata joins)
    FEDSP_Trzcinska_2023_MonKnow__items          name_charset: table name 'FEDSP_Trzcinska_2023_MonKnow' is not lowercase [a-z0-9_.] -- Redivis keeps the case but every client lowercases when joining, so a capitalised name silently drops out of case-sensitive joins.
    FIVPEI_Perrig_2023_AttDiff__items            name is not lowercase (breaks case-sensitive metadata joins)
    FIVPEI_Perrig_2023_AttDiff__items            name_charset: table name 'FIVPEI_Perrig_2023_AttDiff' is not lowercase [a-z0-9_.] -- Redivis keeps the case but every client lowercases when joining, so a capitalised name silently drops out of case-sensitive joins.
    NAMPRB_Siwiak_2024_SSUB__items               name is not lowercase (breaks case-sensitive metadata joins)
    NAMPRB_Siwiak_2024_SSUB__items               name_charset: table name 'NAMPRB_Siwiak_2024_SSUB' is not lowercase [a-z0-9_.] -- Redivis keeps the case but every client lowercases when joining, so a capitalised name silently drops out of case-sensitive joins.
    chile_2023_social-welfare-survey_rr__items   name_charset: table name 'chile_2023_social-welfare-survey_rr' is not lowercase [a-z0-9_.] -- Redivis keeps the case but every client lowercases when joining, so a capitalised name silently drops out of case-sensitive joins.
    chile_2023_social-welfare-survey_ss__items   name_charset: table name 'chile_2023_social-welfare-survey_ss' is not lowercase [a-z0-9_.] -- Redivis keeps the case but every client lowercases when joining, so a capitalised name silently drops out of case-sensitive joins.
    chile_2023_social-welfare-survey_u__items    name_charset: table name 'chile_2023_social-welfare-survey_u' is not lowercase [a-z0-9_.] -- Redivis keeps the case but every client lowercases when joining, so a capitalised name silently drops out of case-sensitive joins.

  not CSV, ignored (1)
    .DS_Store

2 file(s) skipped for errors and will NOT be uploaded:
  gilbert_meta_112__items: dup_item_resp: 5 duplicate item+resp rows with identical option text -- an item text table carries one row per response option, so a repeat is a doubled upload (#1810)
  gilbert_meta_70__items: dup_item_resp: 16 duplicate item+resp rows with identical option text -- an item text table carries one row per response option, so a repeat is a doubled upload (#1810)
  [1/18] FAD_fadplus_goto2021__items -> irw_text_2
  [2/18] FEDSP_Trzcinska_2023_MonKnow__items -> irw_text_2
  [3/18] FIVPEI_Perrig_2023_AttDiff__items -> irw_text_2
  [4/18] NAMPRB_Siwiak_2024_SSUB__items -> irw_text_2
  [5/18] chile_2023_social-welfare-survey_rr__items -> irw_text_2
  [6/18] chile_2023_social-welfare-survey_ss__items -> irw_text_2
  [7/18] chile_2023_social-welfare-survey_u__items -> irw_text_2
  [8/18] estcrm_selfeff__items -> irw_text_2
  [9/18] florida_twins_par__items -> irw_text_2
  [10/18] gilbert_meta_13__items -> irw_text_2
  [11/18] gilbert_meta_14__items -> irw_text_2
  [12/18] gilbert_meta_23__items -> irw_text_2
  [13/18] gilbert_meta_30__items -> irw_text_2
  [14/18] gilbert_meta_45__items -> irw_text_2
  [15/18] gilbert_meta_68__items -> irw_text_2
  [16/18] gilbert_meta_69__items -> irw_text_2
  [17/18] gilbert_meta_71__items -> irw_text_2
  [18/18] wine_luckett2021__items -> irw_text_2

18/18 uploaded and row-count verified.

These are DRAFT versions. Review the diff on Redivis and publish by hand.
```

ben-domingue:
the rest should be public

xingyi-zhang:
@ben-domingue  for the NA,  it came from where option texts are available but it's not related to the response:

ex: 
item_1  …welke soort kracht…?   Duwkracht         resp=NaN     ← option rows
item_1  …welke soort kracht…?   Trekkracht        resp=NaN
item_1  …welke soort kracht…?   Wrijvingskracht   resp=NaN
item_1  …welke soort kracht…?   Zwaartekracht     resp=NaN
item_1  …welke soort kracht…?   option_text=NaN   resp=0       ← resp
item_1  …welke soort kracht…?   option_text=NaN   resp=1


what an ideal format (gilbert_meta_23):
1code  Which number is fifteen?   5    raw_resp=A  correct_response=B  resp=0
1code  Which number is fifteen?   15   raw_resp=B  correct_response=B  resp=1
1code  Which number is fifteen?   50   raw_resp=C  correct_response=B  resp=0

what do we want to do in this situation? do we drop the rows where resp is empty ?

ben-domingue:
Don't drop the empty-`resp` rows — those are the option wording, which for these tables is the only thing the item table adds. It's the other set that's the problem, and the diagnosis is a little different from how you framed it.

## The split isn't the defect

`gilbert_meta_23` — the table you cite as the ideal — has exactly the same split shape, on item `13code`:

```
13code  Check all the numbers that are greater than 3.  option_text=7   resp=NA  raw_resp=A
13code  ...                                            option_text=2   resp=NA  raw_resp=B
13code  ...                                            option_text=4   resp=NA  raw_resp=C
13code  ...                                            option_text=NA  resp=0   raw_resp=NA
13code  ...                                            option_text=NA  resp=1   raw_resp=NA
```

`correct_response=A;C` — a select-all item, so no single option maps to 0/1 and the split is unavoidable. That one is fine, and it's fine for one reason: **every option row is addressable.** `raw_resp` gives each option a code, so a distractor analysis can join to it and the score levels sit alongside.

`gilbert_meta_112`'s option rows carry no `resp`, no `raw_resp` and no `correct_response`. Nothing joins them to anything. That's the defect — not the two row sets, but an option row that asserts wording and no handle on it. The `resp=0/1` rows are then pure gate filler: they carry only the item code, which the response table already has.

So the rule I'd apply: **an option row must carry `resp` or `raw_resp`.** Both blank is a defect. Score-level rows with everything else blank are legitimate only as the residual of a genuinely set-scored or many-to-one item whose options are addressable some other way — `13code`, not `item_1`.

## For 112, in order of preference

1. **Recover the key.** If the deposit's booklet or rubric states the correct answer, fill `correct_response` and map option → `resp` directly: key = 1, each distractor = 0. Many-to-one is normal, and this is the rest of `gilbert_meta_23`. No NA rows, resp set covers {0,1} for free.
2. **Failing that, give the options their printed codes.** If the booklet labels options (A–D, 1–4 — printed MCQs almost always do), that goes in `raw_resp`. The options become addressable, the table looks like `13code`, and `validate_items.R` already accepts `raw_resp` in place of `resp` when the scoring key isn't recoverable (SKILL.md, "resp/option_text"; `gilbert_meta_11` is the precedent).
3. **Only if the booklet numbers nothing and no key exists** do you have unaddressable option text, and the honest shape is to keep the option rows, drop the padding, and say in a `public_note` that option→score is unmapped.

Route 3 is a `FAIL: no resp values at all.` today — there's no documented-gap exit in the gate. That's @ben-domingue's call, not something to pad around; padding is what you get when the only way past a gate is to manufacture rows. My read is the gate should WARN on all-blank `resp` with `option_text` present, since the padded rows are strictly worse than blank: a reader joining on `(item, resp)` lands on `option_text=NA` and concludes there's no wording, and `audit_batch`'s blank rate reads 70% on a file whose options are fully populated.

## Same shape in 203

`gilbert_meta_70` has this too — 8 of 33 items, 16 padding rows, `option_text` 73% blank, and its option rows carry no `raw_resp` either (the Hindi ASER-style items). Neither it nor 112 is live in `irw_text` yet, so both are fixable before upload rather than as corrections. Worth a pass over the rest of the `gilbert_meta_*` family on the same route before either batch goes up.

xingyi-zhang:
Fixed both, your rule applied literally — an option row must carry `resp` or `raw_resp`, both blank is the defect. Each option row now carries **its own wording** in `raw_resp`, so the options are addressable and the score-level rows become the residual rather than padding. No rows dropped.

`gilbert_meta_112` `item_1`, as it now stands:

```
item_1  …welke soort kracht…?  Duwkracht        raw_resp=Duwkracht        resp=NA
item_1  …welke soort kracht…?  Trekkracht       raw_resp=Trekkracht       resp=NA
item_1  …welke soort kracht…?  Wrijvingskracht  raw_resp=Wrijvingskracht  resp=NA
item_1  …welke soort kracht…?  Zwaartekracht    raw_resp=Zwaartekracht    resp=NA
item_1  …welke soort kracht…?  option_text=NA   raw_resp=NA               resp=0
item_1  …welke soort kracht…?  option_text=NA   raw_resp=NA               resp=1
```

`gilbert_meta_112` 7 option rows, `gilbert_meta_70` 24, zero unaddressable in either.

**Why the wording rather than a letter.** Both booklets print bare checkboxes — rendered at 120–150 dpi there is no A–D or 1–4 anywhere on the page, so a letter would be IRW-invented. `gilbert_meta_11` uses letters because its booklet prints them and its `correct_response` says `"A;C"`, i.e. the letters exist to give `correct_response` a vocabulary. There's no `correct_response` here, so nothing needs that vocabulary and the wording is the handle directly. `option_text` keeps the same string — the schema defines it as the literal option wording and that is what it is.

**One correction to your read, in 112's favour.** You said "`13code`, not `item_1`" — but `item_1` **is** genuinely many-to-one. The booklet shows Vraag 1 as *five figures sharing one option list*, and `Test_Validation.pdf` records the first cluster being "combine[d] into one single item". Five correct answers behind one 0/1 score, so no single option could map to it even with a key. It qualifies for your carve-out on the merits, not just after this fix. `item_5` and 70's comprehension items are ordinary single-answer questions whose key is simply absent.

**No key exists, and I checked all three routes before settling for route 2.** For 70: the deposit's own scoring workbook (`RawTestVariables_UTK_BH.xlsx`) has an "Answer Details" column and it is **empty for all eleven** std12 items; `makedata.do` only `rowtotal`s already-scored 0/1 variables, so scoring happened before deposit; and the response data records accuracy only, never the chosen option, so nothing can be solved from responses. Same for 112, where the answers additionally depend on figures.

Six of 70's eight *are* derivable by reading the printed comprehension passage — the story says the rabbit was आलसी and was resting पेड़ के नीचे, both verbatim options. I didn't. That makes IRW the author of the key, which owes an issues-page line under the 2026-09-03 ruling, and `provenance_vocab.csv` has no `key_source` value for a key read off a printed stimulus (`derived_from_responses` would be false). Your call if you want it; it's six items.

**Still fails the gate, and that's now the only open question.** `validate_items` 2/2 on both. `irw-validate` still reports `dup_item_resp`, because it keys on `(item, resp)` and never reads `raw_resp` — so the data satisfies your rule while the check doesn't implement it. Same NaN-key behaviour as the message bug (`duplicated()` counts NaN == NaN while `groupby` drops it, which is why it says "identical option text" when the labels differ). I have that patch written and tested if you want it: 126 tests pass, 2 of the 3 added fail without it.

Pushed to `xingyi/1945-batch203` (`cc7e1a97`) and `xingyi/1945-batch204` (`f2ad159e`).

One unrelated thing worth flagging: **batch_204 never reached main.** #2185 merged into the batch_203 branch about five seconds after #2184 merged that branch to main, so the 26 files are sitting on `origin/xingyi/1945-batch203` unmerged. Happy to open a fresh PR to main whenever.

ben-domingue:
Closing the loop on the three open threads here.

**batch_204 reached main.** #2227 is merged as 79bb51ab — 21 tables across batches 204 and 205, plus the cherry-picked `gilbert_meta_70` fix and the three `ecps_sahm_2024_*` queue rows.

**The gate patch.** Your diagnosis was right and I re-derived it independently before reading this comment: `duplicated()` treats NaN as equal while `groupby` drops NaN keys, so `conflicting` comes back 0 and control falls to a message that asserts the option text is identical when it is the only thing distinguishing the rows. Tracked as #2232 — please open your patch as a PR against main. It currently blocks `gilbert_meta_70` (16 rows) and `gilbert_meta_112` (5) from `red_up`, so it is on the critical path for those two, not cosmetic.

Note your PR description said only `gilbert_meta_112` still fails the gate. `gilbert_meta_70` fails identically — I ran `irw-validate --profile upload` over all 21 tables on the branch. Worth re-running the gate over the whole batch directory rather than the tables you touched last, since the fix you shipped for one file moved the other into the same state.

**batch_205 is uploaded** — 8 of 9 to the `irw_text_2` draft, row-count verified, stamped in #2231. `preussmattsson_2022_ownership` is held: its extraction is on a 6-code item axis while the response data carries 24, and correct item text for it is already published in `irw_text`. See #2230 — and note that six more of the tables you claimed in #2228 are in the same category.

**Your open question about the six derivable keys** on `gilbert_meta_70` — reading the answers off the printed comprehension passage, which would make IRW the author of the key and has no `key_source` value in `provenance_vocab.csv` — is a call for @ben-domingue, and I have not answered it for him. Flagging it so it does not get lost in a merged thread.

xingyi-zhang:
@ben-domingue — three things, briefly.

**The gate re-run.** Taken. You were right that fixing `gilbert_meta_112` moved `gilbert_meta_70` into the same state and my PR description reported only one of them. Running `irw-validate --profile upload` over the whole batch directory rather than the tables I touched last is now what I do — it is what found the one blocking finding left in #2268 (a 62-character table name on `alcoholhealthwarninglabel_brennan_2022_intentions_postexposure`, which is a property of the live table and not fixable in the item text).

**The patch.** Replied on #2232. Short version: yours is the right key shape and mine is not. I drop the unkeyed rows rather than falling back to `raw_resp`, which unblocks both files but stops enforcing the addressability rule instead of enforcing it on the right key — a row with neither `resp` nor `raw_resp` and distinct `option_text` passes under mine. Reworking before I open it.

**The six derivable keys on `gilbert_meta_70`** — still yours, and still open as far as I know. Restating it so it is findable: the six comprehension items have no published answer key, but the answers are readable off the printed passage. Deriving them would make IRW the author of the key, and `provenance_vocab.csv` has no `key_source` value for that (`source_published` and `derived_from_responses` are the two, and neither fits — nothing was derived *from responses*). Either a new vocabulary value, or a ruling that IRW does not author keys and those six ship with `correct_response` empty. I have not acted on it either way.

ben-domingue:
@xingyi-zhang — answering the six derivable keys on `gilbert_meta_70`, since you restated it to keep it findable.

**IRW does not author answer keys.** Where no key is published, `correct_response` stays empty — including these six, whose answers are readable off the passage. That is a standing rule, not a one-off: it applies to every comprehension set that comes through.

So no new `key_source` value. The vocabulary stays at `source_published` and `derived_from_responses`, and you were right that neither fits — the absence was telling you something. The passage ships with the items, so anyone who wants to derive the keys can, and it is clear they did it, not us.

Ship the six with `correct_response` empty and a note saying no key is published.

On the gate re-run: that is the right habit, and it earned its keep straight away on #2268.

PR #2211, opened 2026-09-16
Rights register: correct the ANES row's scope note

ben-domingue:
Follow-up to #2210, which I got wrong on one point of fact.

That row said the redistribution question "was not put to ANES." **It was.** Ben has since shared the outgoing email: a detailed request to *host item-level responses* from the 2016 Time Series feeling-thermometer batteries (V162095–V162113 and V162310–V162314 post-election, V161086–V161096 pre-election as a second wave; ~4,270 respondents; one row per respondent-item pair, responses only). It named the registration step as a **process** question deliberately separated from the licensing question, and asked whether a grant would extend to the thermometer batteries in other waves.

ANES replied that other studies often *borrow questions* from ANES surveys and that citing ANES makes it fine to use the content for research purposes.

So none of the three things asked — redistribution, registration, other waves — is addressed, and the reply's framing suggests the request was read as a wording request. **Being asked in detail and answering something else is a different record from never being asked**, and the note now says which one happened.

### What does not change

The verdict. `ship_with_note` still stands for **question wording**: citation reserves nothing (batches 092/093) and "for research purposes" with no "only" is the batch_163 shape. `polca_election` is still covered retroactively and nothing shipped needs withdrawing.

### What the row now says plainly

- **Do not ingest ANES response data on this grant.** The 2016 thermometer request is unanswered and needs a second, narrower email that asks the hosting question on its own.
- The Harvard Dataverse CC0 redistribution of the 2016 file, which the outgoing email flagged to ANES as a courtesy, drew no response and remains not authorization.

One line changed.


ben-domingue:
Superseded by #2213, which records Ben's ruling that the ANES reply is the permission — covering responses as well as wording, across public-use waves.

Closing rather than updating because this branch also carries the `sel_marca_2025` rebuild commit, which belongs to #2212 and not here. The branch itself is left alone.

PR #2226, opened 2026-09-18
itemtext: ENEM 2013-2022 and 2024-2025 -- 2,026 items over eleven years (#1709)

mateusmazza:
Item text for **every remaining ENEM year**. 2023 shipped in #1848 / #2180; this is the other eleven, as `batch_enem_<YYYY>` per `EXTRACTION_RULES.md` R9.

**2,026 of 2,035 published items across 2013-2022 and 2024-2025.** The nine absent are every item INEP annulled.

| | |
|---|---|
| years | 2013, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022, 2024, 2025 |
| tables | 44 (11 years x 4 areas) |
| items | 2,026 of 2,035 published |
| gates | `normalize_nulls` / `lint_verification` / `check_provenance` / `audit_batch` all pass |
| `audit_batch` | 44 tables: 11 PASS, 33 WARN, **zero hard-gate failures** |
| translations | none ship (#2179) |
| uploaded | nothing — upload stays a separate manual step |

## Where the text comes from

Where INEP published an accessibility (LARANJA / Ledor) edition, the text is read from it, because INEP wrote verbal descriptions of figures, graphs and tables into that edition. **2013, 2015 and 2016 have no accessibility edition at all**, so those come from the standard AZUL booklet through a new `13_join.py --colour` path that requires exactly one `CO_PROVA` at that colour and reports rather than picks when the colour is ambiguous.

`item` is always `CO_ITEM` from `ITENS_PROVA` at the year's standard prova codes — never positional.

## The position convention is measured, not guessed

Each area's offset is derived as `printed_lo - CO_POSICAO_lo` and recorded in `join_report.json`; the gap filler reads that same number rather than re-deriving it. It comes out 0 for a continuous year, ±45 for 2017's per-area numbering, and for 2016 **CH −45 / CN −45 / LC +90 / MT 0** falls out with no special case — 2016 numbers `CO_POSICAO` in INEP's canonical LC/CH/CN/MT order while printing CH/CN on day 1 and LC/MT on day 2. Proven byte-identical on the years that already passed.

## Verification

Every year is checked against INEP's own printed-position key strings (`TX_GABARITO_{CN,CH,LC,MT}`), letter by letter, in every area of every available booklet. A wrong convention lands near chance (~9/45), so exact agreement over ~45 effectively random 5-way letters is decisive.

All exact **except 2017 LC on LARANJA at 43/45** — and that one is a defect in INEP's own accessibility key string, not here:

- `ITENS_PROVA` gives items 39670 and 29586 **D** and **B** in all five colours, LARANJA included;
- the AZUL and AMARELA key strings both give D and B, 45/45 each;
- the extracted option text is **character-identical** between LARANJA and AZUL for both items, so no reordering could justify E and C.

Three independent sources against one. `correct_response` is unchanged.

## Reproduction was measured

The same items were extracted independently from two different colour booklets with separately embedded font subsets and compared by `CO_ITEM`: **99.97% character agreement, zero substituted letters.**

That check is what found printed page furniture — barcodes, running heads, InDesign source filenames, the anti-fraud watermark, the essay draft page — sitting inside `item_text` / `option_text` in **112 items**. All of it is gone, removed by a reversible post-pass rather than by editing the pinned parser.

## Fonts: four pathologies, all repaired inside the PDF

Repaired by completing `/ToUnicode` in the file, so the pinned parser reads these years like any other and item boundaries cannot drift:

- **2015 / 2016 / 2018** embed Arial as a CID-keyed CFF with no glyph names. 2015 and 2016 needed **no new code** — the 2018 decoder applies unchanged.
- **2021** names glyphs by id in `/Differences`. The CMap is completed, never `/Differences` (patching that corrupts every lowercase `l` and `x`).
- **2021 / 2022** set notation in opaque SymbolMT subsets, resolved over the Adobe Symbol encoding.

For 2021, **545 of 770** surviving markers decoded that way, and the result self-checks: CN 88403 now reads `E° = -3,05 V ... I2 + 2 e- -> 2 I-`, where the booklet's own prose two lines earlier says *"potencial de redução igual a menos 3,05 volts"*.

- **MyriadPro** in 2015 had ten CIDs above its ASCII run that the decoder left as `U+FFFD` "rather than guessed". That default was worse than it looked: those are not decorations, they are **letters inside words**, and one item shipped reading `"Use \ufffdgua tratada ... recomenda\ufffd\ufffdes quanto \ufffd restri\ufffd\ufffdo"`. All ten are now resolved from word-level evidence (á, à, ã, ç, é, ó, õ, ú, the fi ligature, and one genuine bullet) and that line reads **"Use água tratada e siga sempre as recomendações quanto à restrição"**.

**There is now no `U+FFFD` anywhere in the corpus**, and no control character either.

## Coverage and the deliberate omissions

**Nine items are excluded under R3** — every one annulled by INEP (`TX_GABARITO='X'`), so there is no correct answer to ship. `audit_batch.R` confirms this is right rather than a gap: those years report `n_items_live == n_items_candidate` with `item_set_match=TRUE`, i.e. the live tables exclude them too.

| year | area | item | INEP's reason |
|---|---|---|---|
| 2018 | MT | 30294 | Anulado |
| 2020 | CN | 32352, 44322 | Anulado |
| 2021 | MT | 117674 | Anulado |
| 2022 | MT | 39443 | anulado pedagogicamente |
| 2024 | CN | 61076 | Pedagógico |
| 2025 | CN | 141557, 141774 | Previamente exposto |
| 2025 | MT | 31350 | Previamente exposto |

**37 items ship `option_text = NA`** — the printed options carry no text that could be transcribed. Per the rule in #1848 nothing is generated to fill them: a generated option label would become a category label someone joins to `resp` for distractor analysis. `correct_response` and `resp_raw` keep the options addressable, and **no `option_text` cell is generated anywhere in this batch.**

I checked eight of the 37 against the printed page and the notes say which:

- **5 confirmed drawings** — 2013 CN 15947 and 2021 CN 62293 (circuit diagrams), 2016 CN 97711 (graph sketches), 2018 CN 111411 (chemical structures), 2013 CN 7867 (five graphs). Four of these previously had **no text at all** and were recovered here by locating the option labels geometrically.
- **3 confirmed *not* drawings** — 2015 MT 14712, 2016 MT 60291 and 2016 CN 86572 have **stacked-fraction** options (86572's are `1`, `4/7`, `10/27`, `14/81`, `4/81`). Those are a gap rather than a deliberate omission: they are legible in the source PDF and recoverable by hand. The notes file them separately from the drawings.
- The remaining **29** were classified from the wording of the stem, and the notes say plainly that this was not re-checked page by page. That heuristic is known to misfire — it read 86572 as a figure item because the stem mentions a circuit — so no claim is made about which kind each one is. What is certain for all 29 is that the extractor found no option text and none was invented.

On figure items the stem legitimately carries the figure's **own** labels (axis names, component letters), because those are printed text and no extractor can separate them from prose. 2013 CN 15947 ends `"...está representado em: A Fase Fase"`, where `Fase` labels a wire in the first diagram. The notes say so, so it is not mistaken for a transcription error.

## An anomaly scan, and the bug it found

Twelve shape-based detectors over all 2,026 items (`37_anomalies.py`) — they look for what a bad parse *looks like*, without knowing what the text should say.

**Six found nothing at all**, and those zeros are the structural guarantee: no option letters outside A–E, never more or fewer than one correct row per item, no item with some options `NA` and others not, no case where the key's option is blank while the others have text, no option carrying the next question's marker, no stem repeating itself.

One found a real bug. `one_option_much_longer` flagged option lengths like `[1,1,1,1,346]`: three items had INEP's figure-description block stuck on the end of **option E** instead of moved to the stem. That fix had been declared done in an earlier parser version, and its marker regex had four faults — it could not match the plural headers (`d[oaes]\b` fails on the following "s" in `"Descrição das figuras:"`), it forbade the newline in a header wrapped across a line, its 40-character window was too short, and being case-insensitive and unanchored it also matched ordinary prose (`"descrição do espaço, como em: ..."`). Parser v10 anchors to a line start and requires the capital; it matches **all 416** blocks the old regex found, misses none, and finds 26 more. 2018 CN 39342, 2018 MT 82123 and 2020 MT 59546 now have clean options.

The tell was never the marker — it was the option-length shape.

One detector fires on **1285 of 2026 items** and is simply wrong: "stem does not end in punctuation" is ENEM's house style, where the option completes the sentence (`"...o circuito equivalente a esse sistema é"`). Noted so nobody mistakes it for truncation.

## Known, disclosed, not blocking

- 2015 MT 14712 and 2016 MT 60291 ship `NA` option text because their options are **stacked fractions**, not drawings — disclosed as a gap, not as a figure item.
- 2021 CN 117867 keeps two raw `[UNDECODED-g34]` markers in `option_text`, where generated text is not permitted.
- 2020 day 1 `instructions` is thin because that cover page extracts to 195 characters at source.

## What the 33 WARNs are

Two classes, both already accepted for 2023: blank `option_text` on the bare-figure items above, and LC row-count anomalies. No WARN is a hard-gate failure — `item_set_match`, `resp_set_match`, `uses_raw_resp` and `canonical_nulls` are TRUE on all 44 tables and `pct_item_text_missing` is 0 on all 44.

## Not in this PR

Nothing uploads to Redivis; that stays manual and separate.


mateusmazza:
@ben-domingue @santiagorq2405 

Agents built this PR after multiple rounds of text extraction, machine revisions, my manual eyeballing and feedback, etc. I caught a bunch of small errors across the board, but the sheer amount of text opens up a lot of possibilities for error. 

If you can, it'd be good to have you guys take a look, ask your agents for clever ways to check for mistakes and inconsistencies, and you can send feedback or ask the agent to edit stuff themselves. 

I think we are close to a shippable version but it would be good to have more rounds of revisions here!

ben-domingue:
Thanks, Mateus. This is careful work. I ran some independent checks, and a few things turned up that the anomaly scan didn't catch:

1. **2018 CN: the minus sign decodes as `þ`** in 111630, 39342, 46621 and 76034 (e.g. `iguais a þ0,44 volt`, `10 þ2 g cm þ3`, `Iþ`). It looks like one bad entry in the 2018 CMap. Could you check whether 2015/2016, which share that decoder, could be affected?
2. **Other wrong glyphs:** 2015 CN 82658 has λ as `Ȝ` (stem and option); 2016 CN 40112 has `ĸ` for the equilibrium arrow; 2022 CN 85445, CN 44969 and MT 63646 still contain private-use symbol-font characters (U+F072/F02B/F02D/F0DE).
3. **Drawing options shipping labels instead of `NA`:** 12 items have duplicate options, e.g. 2013 MT 43001 (`A B`×5), 2013 CN 42570 (`O`×5), 2015 MT 41659 and 2013 CN 33532. Under the #1848 rule, shouldn't these be `NA` like the other 37?
4. **Stacked fractions and exponents:** 14712 and 60291 are `NA`, but 8 others ship newline-split fractions (2013 MT 51219 `17\n70`; 2013 MT 31416 `626 − 106` for 62⁶ − 10⁶), and a few 2013/15/16 stems lose exponents (`6 × 1023`). Could these be treated one consistent way, or at least disclosed?

Minor: `ﬁ` ligatures only in 2017 (~140 items), which is worth normalizing; four numeric option sets out of order (2013 MT 44323, 2015 MT 26277, 2016 MT 83906/95265) that are probably INEP's own, but worth a look.

Checks that found these, if useful for the next round: counting unusual characters by year; flagging identical options within an item; having a model answer a sample of items from the extracted text and comparing with the key.

ben-domingue:
Checked this against the data rather than the description -- pulled both versions of the 18 changed tables and re-derived the claims. The correctness of the fixes I'm satisfied with:

- **Thorn.** Exactly 21 occurrences across exactly 4 items (111630, 39342, 46621, 76034), all cleared. The `17_decode_2018.py` note makes the staleness diagnosis convincing: the CID 237 -> U+2212 override is there with its evidence, so the shipped glyph was a stale artifact and not a decoder bug.
- **2022's duplicated letters.** The all-or-nothing argument isn't just asserted, it's visible in the distribution: in the old data 2022 had 176 items with all five options self-prefixed and **zero** with one to four, and no other year had a single item with >=2. After the fix, 880 letters are gone, nothing anywhere has >=2, and the handful of single hits that remain (4 in 2013, 2 in 2017, one each in 2018/2019/2021/2022) are the article -- "A força sobre a carga positiva...". That's the right call and it's cleanly bounded.
- **The character census balances.** 40 stray apostrophes -> 40 curly quotes; ĸ and → each drop 25 while ⇌ gains 25; Ȝ 10 -> λ 10; F02B/F02D/F0DE/F072 -> + − ⇒ ρ with matching counts. ﬁ goes 10,714 cells -> 0. Row counts, column sets and item keys are identical across all 18 tables; the four lost newlines are item 89637 in 2022 MT, whose prefixes were `"B \n"` rather than `"B "`. `_translated` cells are literal `NA` throughout, so nothing went stale behind the fix.
- **It generalises.** I swept all 48 ENEM item tables at the branch tip, including 2014/2020/2023/2024/2025, which this commit doesn't touch: no private-use characters, no ligatures, no thorn/kra/yogh, no residual letter prefixes. What's left is bullets and a legitimate Ω.

Flagging F072 -> ρ as the one I'd want a second look at eventually, but shipping the spec reading with the uncertainty written down is the right default.

What I can't reproduce is the process half of the commit message. It describes `41_staleness.py`, `42_rebuild.py`, the post-passes as pipeline steps, and nine of eleven years rebuilding byte-identical -- but no script changed in this commit (`git diff 8eb92fde 185814d8 -- '**/*.py'` is empty), and neither file exists anywhere on the branch. The `OPT_INLINE` regex that keeps the second letter is unchanged at `batch_enem_2022/scripts/12_parse_booklet_pdf.py:61`, in all eleven copies.

Which leaves the root cause the commit correctly diagnoses -- corrections living outside the pipeline -- still true of the commit itself: 11,731 hand-corrected CSV lines, and no way to re-run the rebuild that's offered as the evidence.

Two questions before I merge:

1. Are the driver scripts uncommitted, or on another branch? If they exist, I'd like them in the PR; the byte-identical result is the most useful thing here and right now nobody else can reproduce it.
2. Does the next `31_assemble_batch.py` run revert these 880 strips? The message says the recopy already did this once. If the passes aren't in the pipeline yet, what stops it happening a third time?

Related: the 16 scripts are duplicated verbatim across all 11 batch directories, so any pipeline fix has to land eleven times. Probably worth a shared module before the next year's batch rather than after.

mateusmazza:
Working in it!

ben-domingue:
## Sampling the script markup: one confirmed defect in `48_mark_scripts`, plus a consistency lead

Ran trap 50's own prescription — sample the output — over the committed
`batch_enem_*/**__items.csv` on this branch (2013-2022; 2023-2025 carry no script
spans, per R1/trap 51, which checks out). The rules you kept are holding: **zero**
lowercase mid-word splits (`conf_orme`), **zero** uppercase word splits, **zero**
un-collapsed adjacent runs. `H_3O^+` is correct, not a false positive.

Two things did survive, and they are the same rule failing in both directions.

### 1. Confirmed: a spurious `^` marks the element after a subscript

7 tokens, in 2016 / 2018 / 2019:

| year | shipped | should be |
|---|---|---|
| 2018, 2019 | `C_6^H12O6` | `C_6H_12O_6` (glucose) |
| 2016 | `H_2^O` | `H_2O` |
| 2016 | `C_2^H2 (g) → C_6^H6 (l)` | `C_2H_2 → C_6H_6` |
| 2016 | `CH_3^OH` | `CH_3OH` |
| 2016 | `H_2^PO4` | `H_2PO_4` |

Context, 2016 CN: `Fenol C6H5^OH + H_2^O ⇌ C^6H5O− + H^3O+` — where `C^6H5O−`
and `H^3O+` should be `C_6H_5O−` and `H_3O+`.

Diagnosis: the surviving rule is *"a script must change case **or follow a
digit/symbol**"*. In `H_2O`, the `O` follows the digit `2`, so the second clause
fires and marks it as a superscript. This asserts wrong chemistry in chemistry
items, and — trap 50's point exactly — it still reads as text, so nothing gates it.

### 2. Lead, not a finding: trailing runs left unmarked

The converse of the same rule — a digit that follows a letter with no case change
is not marked: `H_2O2` (2015, 2019, 2022), `Fe_2O3` (2017, 2022), `Fe_3O4` (2017),
`Cr_2O7` (2017).

Widening it finds formulas marked inconsistently **within a single line**. 2013:

    FeS_2 (s) + 15 O_2 (g) + 2 H2O (l) → 2 Fe2(SO_4)3 (aq) + 2 H_2SO4 (aq)

`O_2` and `SO_4` marked; `H2O`, `Fe2` and the trailing `3` not. ~67 distinct tokens
of this shape across the nine years.

I am flagging this as a lead rather than a defect on purpose: by trap 51's own
logic the answer depends on what the source PDF encoded as a span, and an unmarked
`H2O` may faithfully reflect a booklet with no subscript there. **Checking one span
in the 2013 source settles the entire class** — if the PDF has it, this is the
same bug as §1; if it doesn't, the output is right and only §1 needs fixing.

### Reproduction

    grep -oP '[^\W\d_]+_\d+\^[^\W\d_]' <items.csv>        # §1
    grep -oP '\b[A-Z][a-z]?_\d+[A-Z][a-z]?\d+\b' <items.csv>  # §2

Not proposing a patch — the fix is a change to your rule, and you have the source
PDFs to check §2 against. Everything else in the branch sampled clean.

mateusmazza:
Quick status update @ben-domingue, there's some stuff still running in the cluster so I will log off for today and pick it up again later in the week. 

**Done:** §1 and §2 both fixed — checked your §2 lead at source first, and the
2013 PDF does encode every digit on that line as a real subscript, so it was a
defect rather than a faithful rendering. Root cause turned out to be the body-size
reference, not the case rule: a 12pt reaction arrow made ordinary 9.75pt text fail
the script test. Your two greps go 50 → 0 and 50 → 0, 644 markers restored,
`H_3O^+` intact, trap 50's shapes still 0, and no text changed — only marker
placement. Pushed at `246a83cc`.

Two other defect classes landed in the same commit, both found by a 100-item
model-answering check: stacked fractions flattened by reading order
(`AC = 7/5 BD` → `AC = 7 BD5`), and figure descriptions attached to the wrong item.

**Still running:** `audit_batch` across all 44 tables (~2.5h, reads the 45M-row
response CSVs). It's auditing the pre-sync tables, so I'll diff its verdicts
against the previous run.

**Next:** finish that audit, then a proper writeup of the two new classes and
the rules behind them.

mateusmazza:
@ben-domingue see the agent's reasoning and comments below. We went through a few additional rounds of iteration to fix subscripts and superscripts, and used that 'model sample test' to see if the possibly sketchy items are answerable by the LLM. Everything looked fine, except the few cases where drawing are not describable and answer options are NA. Let me know if you think any further checks are needed!

-----
Both your §1 and §2 are fixed, and chasing them turned up three more classes
of the same shape. Everything below is at `667b1fcc`.

## §2 is a defect, and your one-span test is what settled it

The 2013 source encodes **every** digit on that line as a genuine 5.22pt
lowered subscript span, including all four we shipped bare:

```
'4 FeS'      9.00      '2' 5.22 SMALL lowered
' (g) + 2 H' 9.00      '2' 5.22 SMALL lowered   <- shipped as H2O
'(SO'        9.00      '4' 5.22 SMALL lowered
')'          9.00      '3' 5.22 SMALL lowered   <- shipped bare
```

## The cause is not the clause either of us suspected

Not *"changes case or follows a digit/symbol"*. Three separate mechanisms:

**§1** — `body` was `max(span size)`. On `C6H5OH + H2O →` the reaction arrow
is set at 12pt against 9.75pt text, so `body` came out 12 and the script test
`size >= body*0.85` classified the **ordinary text** as script. The body text
also sits above the subscripts' bottom edge, which was serving as the
baseline, so it scored as *raised*. That is `H_2^O`. On the `⇌` lines a 12pt
kra poisons the line the same way, which is why `C6H5O−` came out `C^6H5O−`
with the **digit** flipped too. One glyph mis-sizing a whole line, in both
directions — your instinct was right even though the mechanism wasn't the
rule.

Body size is now the **character-weighted mode** of the line, and the baseline
comes only from body-size spans.

**§2a** — `CTX = 24` cut the left context for `H2O`'s subscript to
`"S2 (s) + 15 O2 (g) + 2 H"`, whose first character lands inside `FeS`. The
start-boundary guard — the one added for `T_ANTOS` — then rejected a valid,
24-character-anchored span. It now applies only to untruncated contexts.

**§2b** — the dedupe collapsed any same-kind mark within 6 characters, eating
the second subscript of `(SO_4)_3` and `H_2SO_4`.

| check | before | after |
|---|---|---|
| §1 spurious `^` after subscript | 50 | **0** |
| §2 unmarked trailing digits | 50 | **0** |
| trap 50's three shapes | 0/0/0 | 0/0/0 |
| `H_3O^+` intact | 10 | 10 |
| markers | 1 599 | 2 484 |

`0 cells differ once every _ and ^ is removed` — marker placement moved, no
character of text did. 24/24 sampled new markers map to real small spans.

## Three more classes, same shape

Found by taking your third suggestion further: 100 items, risk-weighted by how
much our pipeline had touched them, answered from the extracted text alone.

**Stacked fractions.** 2013 MT 43849's `AC = 7/5 BD` shipped as `AC = 7 BD5`.
Every character genuine, arithmetic gone. The superscript pass cannot see
these by construction — a stacked numerator is at *full body size* — so the
new pass measures the **bar**, which is a drawing, not text. It declines and
reports rather than guessing, which is what keeps `ABO` from becoming `AB/O`
and `Teste 1:` from becoming `Teste/1`.

**Misplaced figure descriptions.** Your accessibility editions sometimes park
a description at the foot of a page, after the last item's options, describing
an item printed earlier or on the next one. 2018 CN 59858 asks how much energy
oxidising glucose releases and ended with a description of an electrical
circuit. All three had an identifiable owner that carried no description of
its own, so they were **moved**, not stripped.

**Dropped tables.** Same editions, opposite failure: the sentence pointing at
the table survives — *"O quadro apresenta a potência aproximada de
equipamentos elétricos"* — and the numbers don't. Four items, all 2018,
recovered from your standard booklet. Worth saying that scoping this was most
of the work: 47 of 51 candidates are false positives, because "quadro" is an
ordinary Portuguese word, a Dias Gomes theatre scene and a comic-strip panel
before it is a table.

And one I'd have missed without the check: 2013 CN 23920 shipped a stem
reading `6 × 10^23` beside options reading `1,5 × 1025`, because the context
begins with the printed option letter that the parser had already stripped.

## On the check itself

Re-answering the 73 items this work changed: **67 of the 70 answerable
correct**. None of the three misses is a text defect — two are figure-
dependent, one was my own reasoning. I'd still read this as "the text carries
its meaning", not as a score: same model family, not independent.

Worth flagging that answerability is not really the bar here. IRW item text is
read for linguistic complexity, so a figure nobody described costs little,
while a lost table or a spliced-in paragraph changes what is being measured.
That is why the dropped tables got fixed and the undescribed figures did not.

## Things I got wrong on the way, for the record

Trap 50 is about this pass shipping bugs silently, and it happened again:

- my new dedupe didn't advance its cursor on a collapse, so a three-span run
  kept its third span — caught by reading my own diff, not by a test;
- exact abutment missed **overlapping** marks: `Ce4+` arrives as a 2-char span
  `4+` plus a 1-char `+`, and shipped `Ce^4^+`. The suite caught that one. The
  rule that survives is an end cursor, with five shapes unit-tested;
- round-tripping through Python's csv writer rewrote all 48 tables — 24 000
  diff lines — because these are written by R (character columns quoted,
  numeric and NA bare, header always quoted);
- `31_assemble_batch.py`'s copy source still pointed at `out_v8` and bare year
  roots, so running it would have reverted all of the above for the third
  time. To your earlier question about what stops it happening again: `out_rb`
  is now the canonical source, because that is what `41_staleness.py` compares
  against. The refusal gate alone would not have caught it — a reverted
  fraction or subscript leaves text that reads perfectly well;
- a length ratio called 47 of 128 items "short" before I cut the option block
  from the printed side; the real figure was 6, and those are mostly graph
  axis numbers and source credits.

Rules R12–R14 in `EXTRACTION_RULES.md`, traps 52–67 in `STATUS.md`.

## State

44 tables. `audit_batch` 11 PASS / 33 WARN / **0 ERROR**, unchanged.
`lint_verification` 44 rows clean, `check_provenance` exit 0, 0 control
characters, 0 U+FFFD, 0 page furniture. `41_staleness.py`: committed tables
match a fresh pipeline build exactly.

ben-domingue:
## Ruling on the recovered tables (R2, accessibility editions)

Recovering a table's contents from INEP's standard booklet into an
accessibility-edition item is gap-filling under R2 and is owed no "gerada por
IA" marking. The characters are INEP's own, from INEP's own first application
of the same exam; marking them would put an AI provenance flag on text no
model wrote, which misstates the provenance in the other direction. The four
2018 items in cec811b are correct as shipped.

Scope, so this doesn't generalise past what I've actually looked at: this
covers recovery where (a) the source is the standard booklet for the same exam
and year, and (b) the stem already points at the missing content. Anything
reconstructed from any other source, or inserted where the stem does not call
for it, is not covered and should be raised before shipping.

@mateusmazza please write this into `EXTRACTION_RULES.md` as a named rule with
those two conditions, in the idiom of the rules already there, so the next
person meets it as a rule rather than as four merged items. I'd like that in
this PR rather than a follow-up — I'll merge once it lands. Nothing else is
holding it: the branch is clean and both checks are green.

mateusmazza:
@ben-domingue all good! We wrote a new rule below. Happy to go over more iterations and checks if needed. 

 ```
R15. Text recovered from INEP's own booklet is not generated text — the ruling: gap-filling under R2, owes no inline (gerada por IA) and no description_source=partly_generated, because marking it would put an AI provenance flag on text no model wrote. Both conditions are stated as the gate on the rule: (1) the source is the standard booklet for the same exam and year, at that year's standard_prova_codes; (2) the stem already points at the missing content. Two named exclusions — reconstruction from any other source, and insertion where the stem doesn't call for it — go to R11 before shipping. Closes with the line that keeps it narrow: R15 is not a route around R5, it's the statement that INEP's own text never entered R5's scope.
- Cross-references so the rule is reachable from where someone would actually land: a bullet in R5 (the generated-text rule), a bullet in R11's escalation list, and R14's existing "no marking is owed" sentence now points at R15 for the conditions.
```

ben-domingue:
@mateusmazza i was trying to tidy things up and merged this. was that right? do i need to u/l anything at this point?

mateusmazza:
@ben-domingue Nope, all good! Very happy that we were able to get this done!

I'm already thinking of some analyses to run with this data... will keep you posted. Feel free to ping me if your agents find any inconsistencies or issues with the data, I'll be happy to work through them whenever.

PR #2234, opened 2026-09-19
refresh_live_tables: a shard with no draft aborted the whole refresh (#1828)

ben-domingue:
`refresh_live_tables.py` has been failing silently since the last release, and everything downstream inherited the error.

## The bug

When a shard has no open draft — the normal state between releases — the client raises:

```
NotFoundError: Not found: datapages.irw_text:next
```

The guard meant to treat that as "no draft" tested for `not_found` and `404`. That message contains neither, so it fell through to `raise`, the whole refresh aborted, and `live_tables.csv` was never written.

```python
if version == "next" and ("not_found" in msg or "404" in msg):
```

Fixed by matching the lowercased message, which catches all three spellings.

## What it cost

`live_tables.csv` was stuck at **2026-09-06**: 747 published tables, against **1,353** actually published today. Refreshed here.

The snapshot is what `check_issues_page.R` uses to decide whether a table is live — and a note is only DUE on the public issues page once the table is live. So 606 published tables were being classified as "note written, table not live yet, correctly absent", and the checker reported **44** tables owing the page an entry. On the refreshed snapshot it is **261**.

None of that was visible as a failure. The script prints its own as-of date and the checker says how old it is, which is how the staleness was eventually caught — but nothing fails when the refresh does not run.

## Worth considering separately

The one thing this does not fix: a refresh that dies still leaves the previous snapshot in place, which is indistinguishable from a refresh that ran and found no changes. A `--check` run in CI would turn that into a visible failure; the flag already exists.

Refs #1828.


ben-domingue:
One thing the description leaves out, checked before merge: besides adding ~600 tables, the refreshed snapshot **drops 37 that were `published` on 2026-09-06** (19 draft rows also flip to `published`, which is just the release).

All 37 were probed one at a time with `table(...).exists()` on both shards at `version="current"`, not via `list_tables()`, which under-reports. None exists in either shard; the same probe returns `True` for known-live controls. Every one is an instrument under a rights withdrawal between the two snapshots: SWLS ×6, PSS ×7 (incl. `gilbert_meta_59`, `mhscdc_fried_2020_ps`), BFI/BFI-2 ×10 (incl. the seven `sun_2025_morality_*` facet tables), SDQ ×2, HEXACO/SHS ×3, plus MLQ, BPNSF, AQ, IMI, loneliness and similar.

So the snapshot is right and the removals are real. I'm recording it so nobody later reads a 37-row drop in this diff as the refresh losing tables.

